PR #145, opened 2014-07-03
Handle setups where "_source" is disabled.  Update resets values of fields not present in the update-field-list.

vvaradhan:
Use es.update instead of es.index when performing an update as the former allows partial update of documents whereas the later resets/discards values from the index for the fields that are not part of the update field set, thereby making such updated documents unsearchable using any previously searchable query terms or filter.


aherlihy:
Hi @vvaradhan did you make any progress on this PR?

PR #205, opened 2015-01-26
enable rotating log file

crashbtz:
I don't know if I made a good implementation because I don't know python very well, but the idea is to enable a rotating file in order to have few small files instead of a big one which keeps growing.

This PR is not very flexible but I'll be able to set the "when", "interval" and "backCount" parameters of the TimedRotatingFileHandler in the constants.py

But in a first time, I want your advice about this feature

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -26,6 +26,7 @@ import sys
 import threading
 import time
 import imp
+from logging.handlers import TimedRotatingFileHandler
 from mongo_connector import constants, errors, util
 from mongo_connector.locking_dict import LockingDict
 from mongo_connector.oplog_manager import OplogThread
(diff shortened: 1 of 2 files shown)


thelok:
I would find a feature like this useful...

lovett89:
Thanks for the pull request (and great idea, by the way)! I'd like to accept this PR, but I have a couple of suggestions first:

1. I don't see any reason why all log files shouldn't be rotated, rather than having an option to enable/disable it. Rotating log files can be the new default.
2. There should be options for how frequently log files are rotated (i.e., how many days between rotations) and how many to keep around. I suggest `--logfile-interval` and `--logfiles-kept` as the names for these, respectively, but I'm open to other names as well.

Do these changes sound reasonable to you?

crashbtz:
Hi,
Thanks for your response.

I added an option to enable/disable it just because I didn't want to change the first behaviour but set it as the new default sounds goog to me :)

And add options for the "interval", "kept" and maybe the "when" seems better than add them to the constants.py file.

I'll do it tomorrow

crashbtz:
I made the modifications according to what we said.
As you could probably see, I don't speak english very well (I'm french :smile: )
so don't hesitate to correct and/or rephrase my commentaries

crashbtz:
so what do you think @lovett89 ?

lovett89:
merged in e26c1f40f5fee49be981685e68331e508526e642.

lovett89:
Thanks again for the PR!

crashbtz:
No worries and thank you for your help

PR #243, opened 2015-04-16
Remove fields with invalid values (NaN and Infinity)

jaredkipe:
Per decision on 10gen-labs/mongo-connector#238

### DIFF
--- a/mongo_connector/doc_managers/formatters.py
+++ b/mongo_connector/doc_managers/formatters.py
@@ -110,8 +110,8 @@ class DefaultDocumentFormatter(DocumentFormatter):
 
     def format_document(self, document):
         def _kernel(doc):
-            for key in document:
-                value = document[key]
+            for key in doc:
+                value = doc[key]
                 for new_k, new_v in self.transform_element(key, value):
                     yield new_k, new_v
         return dict(_kernel(document))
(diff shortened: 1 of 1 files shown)


jaredkipe:
I tested this approach with similar code. (as I was a bit afraid that 'not yielding' would stop the loop... yeah I wouldn't consider myself a python developer...)

```python
from math import isnan, isinf

def tval(value):
    if isnan(value):
        raise ValueError("nan")
    elif isinf(value):
        raise ValueError("inf")
    return value

def tel(key, value):
    try:
        nvalue = tval(value)
        yield key, nvalue
    except ValueError as e:
        print "Error with key: {} as {}".format(key, e)

def fdoc(document):
    def _kernel(doc):
        for key in doc:
            value = doc[key]
            for new_k, new_v in tel(key, value):
                yield new_k, new_v
    return dict(_kernel(document))


org = { "key1": 1.0, "key2": float("nan"), "key3": 2.0 , "key4": float("inf") }
print org

new = fdoc(org)
print new
```

out
```
{'key3': 2.0, 'key2': nan, 'key1': 1.0, 'key4': inf}
Error with key: key2 as nan
Error with key: key4 as inf
{'key3': 2.0, 'key1': 1.0}
```

llvtt:
Thanks!

PR #248, opened 2015-05-11
Authenticate when connecting to shards

hhromic:
Hello,
I have a **sharded setup with keyFile authentication** as described in the standard MongoDB documentation. I realised that ```mongo-connector``` does not authenticate connections to the shards (when using the ```--admin-user``` and ```--password``` parameters). For this reason, when I try to use the connector I get "not authorized" errors from the shard servers when using this simple authentication method.

While inspecting the source code I found out that indeed you authenticate the primary connection to ```mongos``` and the connections to the replicasets, but not the shards. I added the relevant snippet of code to support this as well.

I've used my patch successfully for some time now with my setup. You will see two additional commits that cancels each other (regarding removing pymongo < 3.0.0 restriction). I realised that pymongo >= 3.0.0 is not transparently working so I reverted this back. My apologies.

I hope this patch is useful,
Hugo.


llvtt:
merged in 2445b551957dfa72a715d1b4a8c2a965114edeca. Thanks for catching and fixing this!

hhromic:
you are very welcome. Looking forward for the next version including this fix :)

PR #278, opened 2015-07-08
disconnect is deprecared in pymongo 3

banzo:
I just replaced the synonym and it works on my setup. 

Let me know if you have any question or remarks


llvtt:
Thanks for the pull request! mongo-connector doesn't support PyMongo 3 yet. Support for it will come after the release of PyMongo 2.9, which makes the transition easier. `close` has been around since PyMongo 2.1, so I think this change is fine. Thanks!

PR #300, opened 2015-07-30
Modified update function to use the unique-key parameter

gsuresh92:
Modified apply_update function to use the unique-key parameter while
updating documents in solr " Error :-"KeyError: '_id'"


llvtt:
Thank you!

PR #302, opened 2015-08-13
Fix the parameter --batch-size

Nagriar:
Hi,

The parameter --batch-size wasn't forwarded to the doc manager because of the usage of different name. I unified it with the name of the option parameter.

What do you think?

Thanks!

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -745,16 +745,16 @@ def get_config_options():
                     " but no doc manager.")
         else:
             if option.value is not None:
-                bulk_size = option.value[0].get(
-                    'bulkSize', constants.DEFAULT_MAX_BULK)
+                batch_size = option.value[0].get(
+                    'batchSize', constants.DEFAULT_MAX_BULK)
             else:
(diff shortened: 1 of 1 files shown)


llvtt:
There are actually two distinct parameters that can be confusing to tell apart. Only one of them is configurable on the command-line: [`--batch-size`](https://github.com/10gen-labs/mongo-connector/wiki/Configuration-Options#batchsize). `--help` describes it:

> Specify an int to update the --oplog-ts config file with latest position of
> oplog every N documents. By default, the oplog config isn't updated until we've
> read through the entire oplog. You may want more frequent updates if you are at
> risk of falling behind the earliest timestamp in the oplog

Since the 2.0 release of mongo-connector, all configuration options are available for configuration via a config file, but not necessarily as a command-line flag, since mongo-connector already had so many. To set "how many documents are sent in a batch to the target system," you want to configure the [`bulkSize`](https://github.com/10gen-labs/mongo-connector/wiki/Configuration-Options#docmanagersindexbulksize) parameter.

PR #303, opened 2015-08-14
Forward the mongoDB client to the doc manager

Nagriar:
Hi,

What do you think to allow the doc manager to fetch the original document from mongoDB instead of the destination?
In our case, we only index some attributes and modify some others to improve the indexing and reduce the size of the index. It's not possible to retrieve the original document from the search engine but this document is still available in mongoDB.

I'm not sure if it's better to have one shared client or to create one client per doc manager. If you have any feedbacks, I'd be happy to improve it :)

Thanks!

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -43,6 +43,18 @@ _SSL_POLICY_MAP = {
     'required': ssl.CERT_REQUIRED
 }
 
+class ConnKeeper:
+    """ Keep the reference to the mongoDB client used by the connector
+    to make it accessible to the doc managers
+    """
+    def __init__(self):
+        self.conn = None
+
+    def set(self, conn):
+        self.conn = conn
+
+    def get(self):
+        return self.conn
 
 class Connector(threading.Thread):
     """Thread that monitors a replica set or sharded cluster.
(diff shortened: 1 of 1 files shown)


Sebmaster:
I'd need this too, although I temporarily fixed this with just

```diff
diff --git a/mongo_connector/oplog_manager.py b/mongo_connector/oplog_manager.py
index c3fc542..158636b 100755
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -91,6 +91,9 @@ def __init__(self, primary_client, doc_managers,
             err_msg = 'OplogThread: No oplog for thread:'
             LOG.warning('%s %s' % (err_msg, self.primary_connection))
 
+        for dm in self.doc_managers:
+            dm.mongo = self.primary_client
+
     @property
     def fields(self):
         return self._fields
```

aherlihy:
Hello!

This looks pretty good, would you mind rebasing with master and adding tests? I can help you get the test suite running, but any new change should come with tests to make sure nothing breaks it in the future.

Thank you for contributing!

ShaneHarvey:
Closing. It looks like all this adds is a reference to `self.main_conn` in the Connector. This was added as part of #563.

Issue #192, opened 2014-11-18
New gridFS support and ability to feed a parent_id (for parent-child relationships)


Analect:
I see from version 2.0 dev, gridFS is now supported, which is great. Is there a way to handle parent-child relationships, so that I can feed a parent_id as part of "files" meta-data.
So, for instance, in mongodb I have an attachments collection and then the actual file attachments are in gridFS.  In ES, I would have an equivalent 'attachments' and 'files' indices, where the parent_id in files is linked to an _id in attachments?  Thanks. Colum

https://github.com/10gen-labs/mongo-connector/wiki/Usage%20with%20ElasticSearch

Analect:
@lovett89 
Very sorry not to have responded sooner. This slipped through the net. I like your proposed solution, which presumably works also where both parent and child IDs happen to come from the same collection. The fix you propose fits my use case, where they are spread across collections.  As regards your last question ... are you referring to the configuration setting for the index in elasticsearch? I suppose if the info is contained in a DocManager config, then if it can be automated to handle passing that config to ES, then it might be nice. However, if the user has a more complex config in ES already involving 'routing' (http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/grandparents.html#grandparents), would your suggested fix be affected by this, since it potentially introduces two parents to a single child. Anyway, a basic implementation along the lines of what you have suggested would be fabulous to have.

Analect:
@lovett89 
Is there anything further I can do to help ref. the parent-child support ... or have you a pretty good handle on the specification you are considering?

lovett89:
@Analect 
I'm still not decided on what needs to be passed to the Elastic DocManager to correctly support parent-child/routing configurations in Elasticsearch. I think your previous comment was correct in pointing out that this could cause problems with "grandparent" relationships. Perhaps this would be better?

    {"<collection name>": {
          "parent": <parent collection>,
          "parent_id_field": <field name>,
          "routing": <grandparent collection>,
          "routing_id_field": <field name>}}

The use of this dictionary would similar to what I described previously, except it includes space for a grandparent collection and grandparent id. Do you think this will take care of the cases that weren't being dealt with in the previous solution? Is there any danger of this being broken by "great-grandparent" mappings?

Effort on this project was focused more on a 2.0 release than on new features for the last couple weeks; sorry that this hasn't been more of a priority.

PR #308, opened 2015-08-25
Add parent-child relationship support to elasticsearch

### LINKED ISSUE #157: Parent child mappings
Is there any way for mongo-connector to bring in documents and have them add the parent ID?  I made a parent mapping on a collection in elasticsearch but don't see an option for the mongo-connector to push in the documents with the parent ID.

Sebmaster:
This adds basic parent-child support to the elasticsearch doc manager.

We require the mongo connection in there since we can't `get` a doc in update without specifying its parent.

Fixes #157, and I guess partially #192 (no grandparent support).

### DIFF
--- a/mongo_connector/doc_managers/elastic_doc_manager.py
+++ b/mongo_connector/doc_managers/elastic_doc_manager.py
@@ -65,6 +65,7 @@ class DocManager(DocManagerBase):
         if self.auto_commit_interval not in [None, 0]:
             self.run_auto_commit()
         self._formatter = DefaultDocumentFormatter()
+        self.routing = kwargs.get('routing', {})
 
         self.has_attachment_mapping = False
         self.attachment_field = attachment_field
@@ -83,6 +84,22 @@ class DocManager(DocManagerBase):
             # Don't try to add ns and _ts fields back in from doc
             return update_spec
         return super(DocManager, self).apply_update(doc, update_spec)
+
+    def get_parent(self, doc_type, doc):
+        if doc_type in self.routing:
+            parent = doc
+            paths = self.routing[doc_type].get("parentPath", [])
+            if len(paths) == 0:
+                return None
+
+            for path in paths:
+                if path in parent:
+                    parent = parent[path]
+                else:
+                    parent = None
+                    break
+
+            return self._formatter.transform_value(parent)
 
     @wrap_exceptions
     def handle_command(self, doc, namespace, timestamp):
@@ -118,8 +135,27 @@ class DocManager(DocManagerBase):
         """
         self.commit()
         index, doc_type = self._index_and_mapping(namespace)
-        document = self.elastic.get(index=index, doc_type=doc_type,
-                                    id=u(document_id))
+
+        document = None
+        if doc_type in self.routing and 'parentPath' in self.routing[doc_type]:
+            # we can't use get() here and have to do a full search instead
+            # this is due to the fact that elasticsearch needs the parent id to
+            # know where to route the get request; we might not have the parent
+            # id available in our update request though
+            res = self.elastic.search(index=index, doc_type=doc_type,
+                                      body={"query":{"ids":
+                                               {"values":[u(document_id)]}}})
+            if res['hits']['total'] == 1:
+                document = res['hits']['hits'][0]
+        else:
+            document = self.elastic.get(index=index, doc_type=doc_type,
+                                        id=u(document_id))
+
+        if document is None:
+            LOG.info("Could not find document with id '%s' in ES to apply update."
+                     % (document_id))
+            return None
+
         updated = self.apply_update(document['_source'], update_spec)
         # _id is immutable in MongoDB, so won't have changed in update
         updated['_id'] = document['_id']
@@ -137,10 +173,20 @@ class DocManager(DocManagerBase):
             "ns": namespace,
             "_ts": timestamp
         }
-        # Index the source document, using lowercase namespace as index name.
-        self.elastic.index(index=index, doc_type=doc_type,
-                           body=self._formatter.format_document(doc), id=doc_id,
-                           refresh=(self.auto_commit_interval == 0))
+
+        parent = self.get_parent(doc_type, doc)
+
+        if parent is None:
+            # Index the source document, using lowercase namespace as index name.
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=self._formatter.format_document(doc), id=doc_id,
+                               refresh=(self.auto_commit_interval == 0))
+        else:
+            # Index the source document, using lowercase namespace as index name.
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=self._formatter.format_document(doc), id=doc_id,
+                               refresh=(self.auto_commit_interval == 0), parent=parent)
+
         # Index document metadata with original namespace (mixed upper/lower).
         self.elastic.index(index=self.meta_index_name, doc_type=self.meta_type,
                            body=bson.json_util.dumps(metadata), id=doc_id,
@@ -172,6 +218,11 @@ class DocManager(DocManagerBase):
                         "_ts": timestamp
                     }
                 }
+
+                parent = self.get_parent(doc_type, doc)
+                if parent is not None:
+                    document_action["_parent"] = parent
+
                 yield document_action
                 yield document_meta
             if not doc:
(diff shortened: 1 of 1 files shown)


mongo_connector/command_helper.py
```diff
@@ -75,3 +75,8 @@ def map_collection(self, db, coll):
             return tuple(ns.split('.', 1))
         else:
             return None, None
+
+    def revert_namespace(self, namespace):
+        full_type = self.dest_mapping.keys()[self.dest_mapping.values().index(namespace)].split(".", 1)
+        return full_type[0], full_type[1]
```
llvtt: It's probably better to create another map like `dest_mapping` called `dest_mapping_inverse` (or something like that) in `__init__` and use that here, since we're potentially calling this method for every document changed in the oplog.

mongo_connector/doc_managers/elastic_doc_manager.py
```diff
@@ -84,6 +86,23 @@ def apply_update(self, doc, update_spec):
...
+                return None
+
+            for path in paths:
+                if path in parent:
+                    parent = parent[path]
+                else:
+                    parent = None
+                    break
+
+            if parent is not None:
```
llvtt: No need to check this. `self._formatter.transform_value(None)` returns `None`.

mongo_connector/doc_managers/elastic_doc_manager.py
```diff
@@ -118,11 +137,20 @@ def update(self, document_id, update_spec, namespace, timestamp):
...
-        # _id is immutable in MongoDB, so won't have changed in update
-        updated['_id'] = document['_id']
+
+        document = None
+        if doc_type in self.routing:
+            # we need to manually retrieve the parent id from mongo
+            db, collection = self.command_helper.revert_namespace(namespace)
+            updated = self.mongo[db][collection].find_one(document_id)
+            if updated is None:
+                return # if the parent was deleted we can't do anything about it
```
llvtt: I'm not sure I understand this part; this method is called whenever a document is updated. Why are we retrieving the most recent version of the document from MongoDB rather than applying the update as usual? What's the significance of `doc_type in self.routing` being `True`?

mongo_connector/doc_managers/elastic_doc_manager.py
```diff
@@ -137,10 +165,20 @@ def upsert(self, doc, namespace, timestamp):
...
+        if parent is None:
+            # Index the source document, using lowercase namespace as index name.
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=self._formatter.format_document(doc), id=doc_id,
+                               refresh=(self.auto_commit_interval == 0))
+        else:
+            # Index the source document, using lowercase namespace as index name.
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=self._formatter.format_document(doc), id=doc_id,
+                               refresh=(self.auto_commit_interval == 0), parent=parent)
```
llvtt: I'm *fairly sure* that passing `None` in as the `parent` keyword argument here has the same effect as not passing it at all, so I think you can just use the code in the `else` branch here.

mongo_connector/doc_managers/elastic_doc_manager.py
```diff
@@ -225,9 +268,17 @@ def insert_file(self, f, namespace, timestamp):
...
+        parent = self.get_parent(doc_type, doc)
+        if parent is None:
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=doc, id=doc_id,
+                               refresh=(self.auto_commit_interval == 0))
+        else:
+            self.elastic.index(index=index, doc_type=doc_type,
+                               body=doc, id=doc_id,
+                               refresh=(self.auto_commit_interval == 0),
+                               parent=parent)
```
llvtt: Same comment here.

mongo_connector/oplog_manager.py
```diff
@@ -91,6 +91,9 @@ def __init__(self, primary_client, doc_managers,
             err_msg = 'OplogThread: No oplog for thread:'
             LOG.warning('%s %s' % (err_msg, self.primary_connection))
 
+        for dm in self.doc_managers:
+            dm.mongo = self.primary_client
```
llvtt: Add a `mongo` member to `DocManagerBase` set to `None`.

llvtt:
Thanks for the pull request! This feature is certainly very useful. I added a few comments to the changes you've made. I also think that this PR would benefit from some tests as well. See [this page](https://github.com/mongodb-labs/mongo-connector/wiki/Running-the-Tests) for instructions on how to set up and run the tests. Let me know if you need any assistance!

mongo_connector/doc_managers/elastic_doc_manager.py
Sebmaster: Just added a comment for this to the source too for later reference but:

You can't do any operations (`get`, `update`, etc.) in elasticsearch without the full routing to a document, and that includes the parent key. Since the delta document we get from the oplog might not contain the parent id, we have to retrieve the document ourselves.

The `doc_type in self.routing` (added `'parentPath' in self.routing[doc_type]` just now too) signifies that this is a parent-child relationship.

We could do a `_search` for the document instead of the Mongo request, not sure which would be more preferable.

mongo_connector/doc_managers/elastic_doc_manager.py
Sebmaster: Nope, that leads to an error like `OperationFailed: TransportError(400, u"ElasticsearchIllegalArgumentException[Can't specify parent if no parent field has been configured]"` when parents aren't enabled.

mongo_connector/command_helper.py
```diff
@@ -25,6 +25,8 @@ class CommandHelper(object):
     def __init__(self, namespace_set=[], dest_mapping={}):
         self.namespace_set = namespace_set
         self.dest_mapping = dest_mapping
+        self.dest_mapping_inverse = dict(
+            (y,x) for x,y in self.dest_mapping.iteritems())
```
llvtt: `iteritems` doesn't exist in Python 3, so this has to be something like:
```python
self.dest_mapping_inverse = dict(
    (self.dest_mapping[k], k) for k in self.dest_mapping)
```

mongo_connector/doc_managers/elastic_doc_manager.py
```diff
@@ -118,11 +136,23 @@ def update(self, document_id, update_spec, namespace, timestamp):
...
-        updated['_id'] = document['_id']
+
+        document = None
+        if doc_type in self.routing and 'parentPath' in self.routing[doc_type]:
+            # we need to manually retrieve the parent id from mongo
+            # this is due to the fact that elasticsearch needs the parent id to
+            # know where to route the get request; we might not have the parent
+            # id available in our update request though
+            db, collection = self.command_helper.revert_namespace(namespace)
+            updated = self.mongo[db][collection].find_one(document_id)
```
llvtt: I see, good explanation. However, I'm not sure that it's necessary to treat documents differently here if they have a parent/routing. Won't the document as stored in Elasticsearch (and retrieved by `get()` below) also have this information?

llvtt:
Thanks for the explanation about needing to retrieve the document to get routing information. I'm not sure that the code needs to be different in the `update()` method, though-- see my comment there. Thanks again for your hard work on this!

mongo_connector/doc_managers/elastic_doc_manager.py
Sebmaster: Yes, it does, but you can't `get()` it without already having this information (the parent).

Sebmaster:
So when I tried running the test suite, I first needed to fix a version conflict with mongo-orchestrator to get it to install properly, however I'm still not able to execute the tests, `python -m unittest tests.test_rollbacks` always errors out with 
```
  File "/usr/lib/python2.7/unittest/loader.py", line 102, in loadTestsFromName
    parent, obj = obj, getattr(obj, part)
AttributeError: 'module' object has no attribute 'test_gridfs_file'
```

llvtt:
@Sebmaster Sorry for the delayed response. I'll refer you to this thread for issues running the tests: https://github.com/mongodb-labs/mongo-connector/issues/311

Sebmaster:
Reworked this to no longer use a `_mongo` attribute in document managers since this caused a whole lot of issues with shards.

> Also related to #303 (the first commit is basically the same, not sure which one you want in here).

therefore no longer applies.

Haven't gotten to write tests yet, we're changing a lot in mongo-connector to make it actually suitable for reeeeaally large indicies.

aherlihy:
Hi @Sebmaster have you made any progress with this PR?

aherlihy:
Closing because the elastic doc managers now live in a separate repo. See https://github.com/mongodb-labs/elastic2-doc-manager/pull/3

Sebmaster:
Sadly trying to keep the dataset consistent slowed down the indexing rate too much to keep up with our ingestion rate, so we had to roll our own with some specific optimizations :(

aherlihy:
When you say "roll our own", did you write your own doc manager?

Sebmaster:
No, we tried that first, but (iirc now) that still barely kept up with ingestion rate so we'd have had to add auto-forking per shard, etc. Additionally, we wanted to add self-healing in case an update was missed or something, so we always pull the latest version from mongo (which we'd have to do anyways for parent-child relationships) instead of converting the oplog to an ES update.

So we wrote a completely new library (which we should open-source soon, it's just not quite ready yet and is deprioritized because it works for us).

aherlihy:
Ah now I understand, and I'm happy it works for you! Let me know when you open source your project :)

PR #309, opened 2015-08-26
Add read preference option for mongo

Sebmaster:
In order to reduce the load on the primaries having the secondaries take over the oplog cursor load is a nice to have.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,6 +3,7 @@
     "__comment__": "To enable them, remove the preceding '__'",
 
     "mainAddress": "localhost:27017",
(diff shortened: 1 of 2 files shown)


llvtt:
Thanks for the patch! I don't it's necessary to have a new option for this, however. This can already be achieved by passing in the `readPreference` argument as part of the MongoDB connection string to `-m`: http://docs.mongodb.org/manual/reference/connection-string/#uri.readPreference

Sebmaster:
Oh sweet, didn't realize that was a thing!

PR #313, opened 2015-09-01
Make logging format configurable

Sebmaster:
This makes the logging format much easier to configure, which makes it easier to switch it to something which can be parsed with logstash for example.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -465,6 +465,9 @@ def get_config_options():
                 "You cannot specify more than one logging method "
                 "simultaneously. Please choose the logging method you "
                 "prefer. ")
+        if cli_values['log_format']:
+            option.value['format'] = cli_values['log_format']
+
         if cli_values['logfile']:
             when = cli_values['logfile_when']
             interval = cli_values['logfile_interval']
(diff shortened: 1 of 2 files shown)


mongo_connector/connector.py
```diff
@@ -520,6 +524,11 @@ def apply_logging(option, cli_values):
         '--stdout', dest='stdout', action='store_true', help=
         'Log all output to STDOUT rather than a logfile.')
 
+    logging.add_cli(
+        "--logformat", dest="log_format", help=
+        "Define a specific format for the log file. This is based on the python logging lib. "
+        "Available parameters can be found at https://docs.python.org/2/library/logging.html#logrecord-attributes")
```
llvtt: Try to keep lines under 80 characters if possible.

mongo_connector/connector.py
```diff
@@ -520,6 +524,11 @@ def apply_logging(option, cli_values):
         '--stdout', dest='stdout', action='store_true', help=
         'Log all output to STDOUT rather than a logfile.')
 
+    logging.add_cli(
+        "--logformat", dest="log_format", help=
```
llvtt: Most of the command-line options have dashes that separate words, so I think this should be `--log-format`.

Sebmaster:
Fixed

llvtt:
Merged in e68c0710341d008963d0455339a501b9bc0d17e1. Thanks for the pull request!

PR #321, opened 2015-09-04
Fix replica set connections

### LINKED ISSUE #314: Reading from secondary still read from primary
Hi !

It seems like the mongo connector is reading from the primary , even if we specify that we would like to read on the secondary with a specific tag, and cause some performance issue on the main application.

Here how looks like the connection string
````
mongodb://MONGOS_SECONDARY_STRING:10000/?readPreference=secondary&readPreferenceTags=use:backup
````

I noticed also in the progress file and the debug the mention of the secondary and the primary 
````
MongoConnector: Starting connection thread MongoClient([u'SECONDARY:10004', u'PRIMARY:10004'])
````

Is it a normal behavior? I was wondering if I missed anything to make sure that it only connect to secondary?

Thank you

Sebmaster:
A combination of not forwarding the connection string to sharded instances + using the wrong pymongo class to connect to mongo makes it impossible to properly provide read preferences to mongo.

MongoClient (in pymongo 2.x) assumes that the connected mongo instance is a standalone instance, which doesn't matter for the initial connection to the mongos/discovery instance, but is problematic for the replica set connection.

When specifying a read preference, it doesn't apply to the MongoClient connection, since they're standalone connections. If we were connected via mongos, that'd be no problem, since mongos would handle this, but we only use the mongos connection to get to the real instances and the standalone client just ignores the read preference.

By using MongoReplicaSetClient, we ensure that the pymongo properly throws on read preference errors and also properly switches between instances.

This should fix #314.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -309,6 +309,7 @@ class Connector(threading.Thread):
                 time.sleep(1)
 
         else:       # sharded cluster
+            queryStr = self.address.split('?')[1]
             while self.can_run is True:
 
                 for shard_doc in main_conn['config']['shards'].find():
@@ -337,8 +338,9 @@ class Connector(threading.Thread):
                             dm.stop()
                         return
 
+                    address = 'mongodb://' + hosts + '/?' + queryStr
                     shard_conn = MongoClient(
-                        hosts, replicaSet=repl_set, tz_aware=self.tz_aware,
+                        address, replicaSet=repl_set, tz_aware=self.tz_aware,
                         **self.ssl_kwargs)
                     if self.auth_key is not None:
                         shard_conn['admin'].authenticate(self.auth_username, self.auth_key)
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -34,6 +34,7 @@ from mongo_connector.command_helper import CommandHelper
 from mongo_connector.util import log_fatal_exceptions
 
 from pymongo import MongoClient
+from pymongo import MongoReplicaSetClient
 
 LOG = logging.getLogger(__name__)
 
@@ -280,7 +281,7 @@ class Connector(threading.Thread):
 
             # Establish a connection to the replica set as a whole
             main_conn.close()
-            main_conn = MongoClient(
+            main_conn = MongoReplicaSetClient(
                 self.address, replicaSet=is_master['setName'],
                 tz_aware=self.tz_aware, **self.ssl_kwargs)
             if self.auth_key is not None:
@@ -339,7 +340,7 @@ class Connector(threading.Thread):
                         return
 
                     address = 'mongodb://' + hosts + '/?' + queryStr
-                    shard_conn = MongoClient(
+                    shard_conn = MongoReplicaSetClient(
                         address, replicaSet=repl_set, tz_aware=self.tz_aware,
                         **self.ssl_kwargs)
                     if self.auth_key is not None:


pjaffe:
I just started my initial proof of concept work today to mirror a large sharded cluster of replicasets to a single mongod that I'll then be targeting for reporting.  I've crossed my fingers that mongo-connector would be the magic bullet for that, and apparently I hit this problem out of the gate... It seems completely ironic that for a project whose first release was a year and a half ago, that the problem I hit was just raised in the last 2 weeks, and I'm thrilled to see a pull request out there to fix this...

Any chance a new PyPi version is expected soon that contains this fix?  Or should I build from source off this branch in the meantime?

Sebmaster:
@pjaffe What's your current testing setup? We've currently tested this with sharded replica-sets behind mongos (specify mongos host in this case) and a single-instance replica set (a.k.a. me testing locally).

pjaffe:
Before noticing your comment (thanks for the quick reply!), I had been revising my original comment above, since I realized it was lacking in some details.

Given your comment here, I'm wondering if the difference is that your replica set has a single instance.  Do you mean that you only have a primary, and no secondaries in your replicaset?  If that is the case, I'm sure that explains why your situation is working.  I would deeply appreciate it if you could throw a secondary into the mix, and see if you can get it to work!

I started to try to set up a development environment for this, but I'm guessing I have a larger barrier to entry than you do at this point.

Thanks again!

Sebmaster:
> Do you mean that you only have a primary, and no secondaries in your replicaset?

Locally yeah, but we also have a staging setup running with this which is a sharded replicaset across multiple instances.

Did you only pull down this branch, or merge in my other changes too? Because ec13f7b looks like it might cause this problem too.

Sebmaster:
Could it be, that your instance on localhost is a secondary? I didn't modify the mongo doc manager, so it won't do auto discovery in case that's a replica set too...

pjaffe:
You may be getting at the crux of my problem there. My target is currently a standalone non-replicaset and non-sharded instance. I had hoped that would work, since it was the simplest conceptual thing, and I assumed would be effectively equivalent to elasticsearch or solr as a target. But I wondered if the mongo doc manager depended on replicaset semantics on the rather as well. But I had assumed that would have been clear in the error logs. My interpretation of my error logs was that it was failing to generate the dump from the "main address" and that it hadn't gotten to the target yet. There was no record in the log of the connection to localhost at all.

Sorry to use you as a sounding board for this if my issue is unrelated. But it still seems closely related to me...

Sebmaster:
No it seems it can actually read all the data and gets stuck at the insert for some reason. However that shouldn't happen with its own connection... I'll try to spin up some instances to test this.

pjaffe:
@Sebmaster - I do now see that the error is on the upsert part of the dump, and not the extract side.  I ran rs.initiate() on the target MongoD, and now a new error occurs, and later in the flow.

So, at this point, this pull request seems to be well intended, and maybe will be something I care about with regard to avoiding negative performance impact on the primaries in my mainAddress.  But the error I encountered and described earlier in this thread now seems unrelated, and was simply due to the target MongoDB needing to be a replicaset.

pjaffe:
I feel bad about polluting this issue's discussion with my previous comments that turned out to be an unrelated user error.  But I am now running mongo-connector and seeing it putting load on the source cluster's primaries while it reads their oplog, instead of mirroring from the secondaries given my inclusion of the ```?readPreference=secondary``` query string in my ```mainAddress``` connection string.

Is there a process for drawing more attention to this pull request and hopefully getting it merged into the master branch and getting a new version released and published to PyPI???

llvtt:
`MongoReplicaSetClient` is deprecated in PyMongo 3, which mongo-connector will soon support. It's actually not even necessary to connect to replica sets. The problem mongo-connector has connecting to secondaries is that the `readPreference` connection string argument isn't sent to the OplogThread. Something like this in connector.py should suffice:

    shard_conn = MongoClient(
        hosts, replicaSet=repl_set, tz_aware=self.tz_aware,
        read_preference=main_conn.read_preference)

Sebmaster:
> The problem mongo-connector has connecting to secondaries is that the readPreference connection string argument isn't sent to the OplogThread. 

It seems like even if provided the readPreference via the connection string, `MongoClient` will not pay attention to it while `MongoReplicaSetClient` will. We tested that out in our server setup - MongoClient will just continue to connect to the primaries.

aherlihy:
Waiting for input on https://github.com/mongodb-labs/mongo-connector/issues/314

aherlihy:
Since this was determined to be a PyMongo version issue I'm going to close this. Feel free to reopen or file a new PR if something else comes up. Thank you for contributing!

PR #323, opened 2015-09-10
should drop mapped db

MartinNowak:


llvtt:
Good catch! Thanks for the pull request.

PR #335, opened 2015-10-09
add emptyCheckInterval option to improve performance

quentins:
In the case where a replicaset doesn't have any docs matching the
target namespaces (which happens with tag aware sharding), we don't
want to keep generating (empty) oplog cursors on that replica set.
This option uses a cheaper `db.collection.count()` call to see if
a replica set has any matching docs and waits a configurable amount
of time before checking again.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,6 +8,7 @@
     "batchSize": -1,
     "verbosity": 0,
     "continueOnError": false,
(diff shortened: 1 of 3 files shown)


aherlihy:
I've restarted the travis build because it seems it failed for unrelated reasons.

PR #356, opened 2015-11-18
change tests use right doc manager, setup id fields appropriately

aganapat:
Support for using a single meta collection
1. Added Support for using a single meta collection which will be a capped collection. This size can be given using the configuration options.
2. Change mongo_doc_manager test case to use the proper doc manager and also support running the tests with single meta collection.
3. As part of using single meta collection, _id field cannot be used in meta collection as there can be duplicates, create a different field (doc_id) which will capture the id of the original document.

### DIFF
--- a/mongo_connector/doc_managers/mongo_doc_manager.py
+++ b/mongo_connector/doc_managers/mongo_doc_manager.py
@@ -61,27 +61,30 @@ class DocManager(DocManagerBase):
         self.namespace_set = kwargs.get("namespace_set")
         self.chunk_size = kwargs.get('chunk_size', constants.DEFAULT_MAX_BULK)
         self.use_single_meta_collection = kwargs.get(
-                                        'use_single_meta_collection',
-                                        False)
+            'use_single_meta_collection',
+            False)
         self.meta_collection_name = kwargs.get(
-                                        'meta_collection_name',
-                                        constants.DEFAULT_META_COLLECTION_NAME)
+            'meta_collection_name',
+            constants.DEFAULT_META_COLLECTION_NAME)
         self.meta_collection_cap_size = kwargs.get(
-                                    'meta_collection_cap_size',
-                                    constants.DEFAULT_META_COLLECTION_CAP_SIZE)
+            'meta_collection_cap_size',
+            constants.DEFAULT_META_COLLECTION_CAP_SIZE)
+
         self.id_field = 'doc_id' if self.use_single_meta_collection else '_id'
+        self.meta_database = self.mongo["__mongo_connector"]
 
         """ create the meta collection as capped if a single meta collection
             is preferred """
         if self.use_single_meta_collection:
-            if self.meta_collection_name not in \
-                    self.mongo["__mongo_connector"].collection_names():
-                self.mongo["__mongo_connector"].create_collection(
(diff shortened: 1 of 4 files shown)


mongo_connector/doc_managers/mongo_doc_manager.py
```diff
@@ -60,10 +60,38 @@ def __init__(self, url, **kwargs):
...
+        self.use_single_meta_collection = kwargs.get(
+                                        'use_single_meta_collection',
+                                        False)
+        self.meta_collection_name = kwargs.get(
+                                        'meta_collection_name',
+                                        constants.DEFAULT_META_COLLECTION_NAME)
+        self.meta_collection_cap_size = kwargs.get(
+                                    'meta_collection_cap_size',
+                                    constants.DEFAULT_META_COLLECTION_CAP_SIZE)
+        self.id_field = 'doc_id' if self.use_single_meta_collection else '_id'
```
llvtt: Why have a separate field called `doc_id` instead of putting this in `_id` as usual?

tests/test_mongo_doc_manager.py
```diff
@@ -35,15 +35,29 @@ def setUpClass(cls):
...
-        cls.choosy_docman = DocManager(
-            cls.standalone.uri,
-            namespace_set=TestMongoDocManager.namespaces_inc
-        )
 
     def setUp(self):
+       
+        """ setup instance variables for running doc manager test
+            using single meta collection or collection per namespace """
+        if type(self) is TestMongoDocManagerWithSingleMetaCollection:
```
llvtt: This seems a little messy. A suggestion: define `use_single_meta_collection` at the class level of `MongoTestCase` and set it to `False`. Below, define your class `TestMongoDocManagerWithSingleMetaCollection` as just setting `use_single_meta_collection` to `True`. When creating the DocManagers below, set `use_single_meta_collection` to `self.use_single_meta_collection`.

llvtt:
I think this looks pretty good. I left a couple of comments. Two more things as well:

1. Could you please format your code to match the style around it? mongo-connector follows [pep8](https://www.python.org/dev/peps/pep-0008/) as closely as reasonable. I.e., 80-character line limits, no continuation lines with `\`, and continuation lines are indented 4 spaces. I can make individual comments for these if you prefer, or I can make a fixup commit after merging this PR if you don't want to deal with it.
2. Please rebase this code on top of the latest commit in master. I fixed the Travis build yesterday, so we should be able to test this PR in Travis once it's rebased.

mongo_connector/doc_managers/mongo_doc_manager.py
aganapat: @llvtt Since we will be writing all of these documents in same collection we need to have the _id auto incremented. Hence the need for doc_id.

tests/test_mongo_doc_manager.py
aganapat: @llovett thanks, will make that change.

aganapat:
@llvtt i have run autopep8 to fix the formatting issues you mentioned.

aganapat:
@llvtt please review, all tests pass and I have handled your comments.

llvtt:
Merged. Thanks for all the hard work!

PR #361, opened 2015-11-24
add sentry to the supported logging methods

oxymor0n:
add the option to direct the log stream to Sentry, an open source log aggregation platform (see [their homepage](https://www.getsentry.com/) and [GitHub repository](https://github.com/getsentry/sentry))

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,6 +8,7 @@
     "batchSize": -1,
     "verbosity": 0,
     "continueOnError": false,
+    "skipCount": false,
 
     "logging": {
         "type": "file",
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -151,6 +151,7 @@ class Connector(threading.Thread):
             collection_dump=(not config['noDump']),
             batch_size=config['batchSize'],
             continue_on_error=config['continueOnError'],
+            skip_count=config['skipCount'],
             auth_username=config['authentication.adminUsername'],
             auth_key=auth_key,
             fields=config['fields'],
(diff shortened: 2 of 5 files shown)


llvtt:
This seems like a cool idea. However, rather than manually providing support for SentryHandler, perhaps we could add an option for the user to specify their own logging configuration, including the specific Handler implementation. This could be added to the config file and then loaded with [`logging.config.dictConfig`](https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig). The only reason this isn't done already is that this feature only exists on Python 2.7+, but mongo-connector supports Python 2.6. However, Python 2.6 users who don't want to upgrade can still use the existing logging configuration options (Sentry won't be available). How does that sound? Any interest in taking this on?

oxymor0n:
That sounds like a nice approach to this problem, since this will make mongo-connector's logging very extensible and standard compliant. I can do it but can you point me to the relevant section of the code?

llvtt:
1. We're creating a new option called "config" inside the "logging" configuration blob, so we need to grab the configuration dictionary out of the config file by checking for "config" in "option" here: https://github.com/mongodb-labs/mongo-connector/blob/master/mongo_connector/connector.py#L459. The config file will end up looking like this:

    ```json
    {
      "logging": {
        "config": {
          "handlers": {
            "console": {
              "class": "logging.StreamHandler",
            }
          }
        }
      }
    }
    ```

    To use this configuration, all that needs to happen is to call `logging.config.dictConfig(options['config'])`. There should also probably be some warning issued if the user tries to specify 'logging.config' and any other log option (e.g., 'logging.filename'), since these options should be mutually exclusive.

2. Add a test in tests/test_config.py for the 'logging.config' option.

Let me if I can do anything else to support your effort. Thanks a lot for doing this!

PR #364, opened 2015-12-02
fix default syslog port (should be 514)

MartinNowak:


MartinNowak:
This was broken by https://github.com/mongodb-labs/mongo-connector/commit/6b8a6649e9008397ef2eb4fe1183541b90e2de5b.

llvtt:
Good catch. Merged.

Issue #226, opened 2015-03-18
--exclude-fields option


llvtt:
Allow users to specify fields within a MongoDB document to exclude.

sjalali:
Observed this behavior on nested fields:
{ "docType": "movie", "searchable": { "title": "Nebraska" }}  (part of much larger document)
In config.json: {"fields": ["searchable", "docType"] }

At startup, initial load, both docType and searchable make it to Elastic.
On subsequent updates, touching docType propagates the change. Touching searchable.title doesn't.
If I change fields to {"fields": ["searchable.title", "docType"] } it works.

Maybe we could have something like "searchable.*"  in fields to specify any changed sub-fields?
Otherwise all sub-fields would need to be listed in the filter.
Thanks!

llvtt:
@sjalali That's a good suggestion! I'll make that into a new feature request (this particular issue is about excluding fields rather than including them).

aherlihy:
https://github.com/mongodb-labs/mongo-connector/commit/4fe5de8d55a6c84d7465bc989f33696144ea9640

Morriaty-The-Murderer:
Was this option still not fulfilled?
My mongo-connector version is 2.4, I used this option in command-line, but failed
```bash
/usr/local/bin/mongo-connector -m 10.10.20.155:27118 -t localhost:9200 -d elastic2_doc_manager -n ask.zd_topic -a usvr_sereader -p XXXXXXXXX --continue-on-error --exclude-fields pic_urls

mongo-connector: error: no such option: --exclude-fields
```

Thank you for help!

PR #369, opened 2015-12-17
Added exclude field

akgoel-mo:
Adds enhancement requested in #226


akgoel-mo:
@llvtt Can this PR be merged? All test cases are passing in python 2.x, except for 1 which is failing for python 3.x and which I am not able to figure out why. As its part of the Solr tests which I didnt even touch.

I need this feature and dont want to maintain my own fork

aherlihy:
Hello! Unfortunately PRs that break the test suite cannot be merged into master. I would recommend rebasing with the current master, which will trigger another travis build. If that doesn't pass then there is a breaking change in your code (or an existing bug) that needs to be addressed. Thank you for contributing!

Issue #316, opened 2015-09-02
Duplicate entries in oplog.timestamp file


XDestination:
I am using the mongo-connector to replicate some data from a 7-shard cluster to a single instance of mongoDB. Today I realized that the number of documents between the two databases is drifting apart and therefore I started to investigate.

The first thing that looks weird to me is that there are duplicate entries in the oplog.timestamp file with different values for the timestamp:

```
[
	["Collection(Database(MongoClient([u'server01.abc.com:27018', u'server02.abc.com:27018', u'server03.abc.com:27018', u'server04.abc.com:27018']), u'local'), u'oplog.rs')", 6187759245908246585], 
	["Collection(Database(MongoClient([u'server05.abc.com:27018', u'server06.abc.com:27018', u'server07.abc.com:27018', u'server08.abc.com:27018']), u'local'), u'oplog.rs')", 6189300503512350739],
	["Collection(Database(MongoClient([u'server05.abc.com:27018', u'server06.abc.com:27018', u'server08.abc.com:27018', u'server07.abc.com:27018']), u'local'), u'oplog.rs')", 6190048532196491279], 
	["Collection(Database(MongoClient([u'server09.abc.com:27018', u'server10.abc.com:27018', u'server11.abc.com:27018', u'server12.abc.com:27018']), u'local'), u'oplog.rs')", 6190048532196491282], 
	["Collection(Database(MongoClient([u'server13.abc.com:27018', u'server14.abc.com:27018', u'server15.abc.com:27018']), u'local'), u'oplog.rs')", 6190048115584663561],
	["Collection(Database(MongoClient([u'server16.abc.com:27018', u'server17.abc.com:27018', u'server18.abc.com:27018', u'server19.abc.com:27018']), u'local'), u'oplog.rs')", 6190048523606556710],
	["Collection(Database(MongoClient([u'server16.abc.com:27018', u'server19.abc.com:27018', u'server18.abc.com:27018', u'server17.abc.com:27018']), u'local'), u'oplog.rs')", 6189298321668964353],
	["Collection(Database(MongoClient([u'server20.abc.com:27018', u'server21.abc.com:27018', u'server22.abc.com:27018', u'server23.abc.com:27018']), u'local'), u'oplog.rs')", 6190048515016622104],
[...]

PR #370, opened 2015-12-17
Use replicaset-name as identifier in oplog-timestamp-file

XDestination:
Currently the list of hosts in a replicaset is used as an identifier in the oplog-timestamp-file. This works well as long as the members in a replica set do not change. If this is the case, mongo-connector gets confused and attempts an initial import (because it is not able to find a proper entry in the oplog-timestamp-file).
This PR provides a fix for that and has legacy-support for the old format included (read old/new name, write new name only).

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -288,7 +288,7 @@ class Connector(threading.Thread):
 
             # non sharded configuration
             oplog = OplogThread(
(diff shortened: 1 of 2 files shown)


XDestination:
This PR is a fix for https://github.com/mongodb-labs/mongo-connector/issues/316

PR #382, opened 2016-01-25
Fix issue with nested properties

robertaistleitner:
There were issues with updating/inserting documents, where nested
properties are not available in the index because they got filtered out
by the oplog manager.

e.g. if you specify to limit to the field "report.tags" of a given
collection, updating works fine when using $set in mongodb. When
inserting a complete new document to the collection the
pop_exclude_fields gets a full document (without dot-notation like with
$set), and behaves incorrect in terms of popping excluded fields (it
looks for "report.tags", but only "report" is in the document which gets
popped.


aherlihy:
Hello, sorry for the delay. It seems your pull request doesn't work for python 3.x, you can see [the Travis CI build here](https://travis-ci.org/mongodb-labs/mongo-connector/builds/104565582). Thank you for contributing!

robertaistleitner:
@aherlihy: can you please help me out:
I don't get the test case 
```

File: mongo_connector/oplog_manager.py [hunk 1]
```diff
@@ -106,8 +106,25 @@ def fields(self, value):
             self._fields = set(value)
             # Always include _id field
             self._fields.add('_id')
+
```
aherlihy: It is possible that the user has only set fields using the constructor, and not this attribute. Anything you do here must also happen in the constructor (line 87) since it's possible that this will never get called, and self._fieldsdict will cause an AttributeError in pop_excluded_fields (line 369)

mongo_connector/oplog_manager.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, primary_client, doc_managers,
         self.continue_on_error = kwargs.get('continue_on_error', False)
 
         # Set of fields to export
-        self._fields = set(kwargs.get('fields', []))
+        self.fields = set(kwargs.get('fields', []))
```
aherlihy: I think this should be self._fields, is there a reason why you're setting self.fields instead?

mongo_connector/oplog_manager.py
```diff
@@ -341,11 +358,25 @@ def filter_oplog_entry(self, entry):
         if not self._fields:
             return entry
 
-        def pop_excluded_fields(doc):
-            # always include _id
-            fields_with_id = self._fields.union(set(['_id']))
-            for key in set(doc) - fields_with_id:
```
aherlihy: Here we're using field_with_id instead of fields because we want to make sure we're never removing _id from the fields that get copied over to the sink.

aherlihy:
Hello!

I've made some comments on your pull request, please let me know if I've misunderstood anything. I made the changes I suggested here: https://github.com/mongodb-labs/mongo-connector/compare/mongodb-labs:master...aherlihy:usersnap-master and it seems to be passing the failing test on my machine. [It's currently running on travis.](https://travis-ci.org/aherlihy/mongo-connector/builds/118225661)

Feel free to neaten up what I've changed, maybe we can pull the code that sets self._fieldsdict into a helper function so that it's not repeated [in the constructor](https://github.com/mongodb-labs/mongo-connector/compare/mongodb-labs:master...aherlihy:usersnap-master#diff-9e5fc676a1f3b77e19871877266b4cf0R88) and [in the fields setter](https://github.com/mongodb-labs/mongo-connector/compare/mongodb-labs:master...aherlihy:usersnap-master#diff-9e5fc676a1f3b77e19871877266b4cf0R126).

The last thing I before I can merge this is a test. If you could add a quick test to [tests/test_oplog_manager.py](https://github.com/mongodb-labs/mongo-connector/blob/master/tests/test_oplog_manager.py) to make sure that your code works and that it doesn't break in the future, that would be great.

Thank you again!

aherlihy:
Also, the test case your failing isn't a weird test case, it is part of fixing https://github.com/mongodb-labs/mongo-connector/issues/409 this bug.

robertaistleitner:
Funny that I haven't seen your last comment at all, just the comments in the code - but I did it exactly the way you described it in the comment ;)

Need to write the additional tests though ;)

aherlihy:
I've been looking through the tests cases and I think I was mistaken at first - I think you may have uncovered some inconsistent behavior (that existed before this PR, leftover from the 3.2 refactoring) between setting fields with the attribute and with the constructor. I'm going to look into it now, thank you for bringing it to my attention!

nVitius:
any updates on the status of this? I would love to have updates to nested properties working.

aherlihy:
Yes, we've decided to rework how the '--fields' option works internally and add support for an '--exclude-fields' option. It is in the final rounds of code review and you can expect a fix within the next day or so. 

After looking into how 'fields' worked previously, and how we expect it to work internally, we decided it would be too expensive and unintuitive for mongo-connector to pop off all the unincluded fields for a user who has large documents and a small number of fields to include. Since we already had a lot of interest in an '--exclude-fields' option, we decided to rewrite '--fields' so that internally it will copy over the included fields to a new document. The '--exclude-fields' argument pops off the excluded fields from the original document.

There are no surface changes to 'fields', and both arguments support nested documents (although not traversing arrays, i.e. doc={'a':[{'b':1}, {'c':1}]}, the path a.1.b would not be supported).

Thank you for your patience!

aherlihy:
Here are the changes I mentioned: https://github.com/mongodb-labs/mongo-connector/commit/4fe5de8d55a6c84d7465bc989f33696144ea9640

aherlihy:
Thank you for contributing! Since the issue has been fixed in a related commit, I'm going to close this pull request.

Issue #397, opened 2016-03-01
TypeError when specify fields in config.json


tad2021:
I want to index some fields of the documents stored in MongoDB. Then I config the config.json in the way below.
...
"fields": ["_id","title","content","author"],
..
Finally, start up connector using command "mongo-connector -c config.json".

Oops, An error jump out suddenly. The traceback is
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.6/site-packages/mongo_connector/util.py", line 85, in wrapped
    func(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/mongo_connector/oplog_manager.py", line 206, in run
    if not self.filter_oplog_entry(entry):
  File "/usr/lib/python2.6/site-packages/mongo_connector/oplog_manager.py", line 351, in filter_oplog_entry
    pop_excluded_fields(entry_o)
  File "/usr/lib/python2.6/site-packages/mongo_connector/oplog_manager.py", line 345, in pop_excluded_fields
    for key in set(doc) - self._fields:
TypeError: unsupported operand type(s) for -: 'set' and 'list'

behackett:
@aherlihy this looks like a bug in 2.2.

aherlihy:
Sure does. Looking into it!

tad2021:
Bug fixed by @aherlihy

PR #398, opened 2016-03-01
Fix up #397

tad2021:
Fix up #397 .
The result of set(doc) is a set, but self._fields is a list, so the error happened.
The solution is transfer this list to set.

for key in set(doc) - set(self._fields):

It works!!!


aherlihy:
Hello!

Thank you for reporting this, it's definitely a bug. We're going to fix it in a slightly different way, and add a test. You can see the fix here: https://github.com/mongodb-labs/mongo-connector/compare/master...aherlihy:v2.2. 

Your contributions are appreciated!

PR #399, opened 2016-03-01
Use replicaset-name as identifier in oplog-timestamp-file

XDestination:


aherlihy:
Hello! Sorry for the delay in getting back to you.

Thank you for identifying a bug and coming up with a solution.

At first glance it seems your changes have broken the existing test suite (see here https://travis-ci.org/mongodb-labs/mongo-connector/builds/112793594). Any changes to mongo-connector should pass the test suite, but if that is not possible, fixes for the tests should be included in the pull request. Any breaking changes should be addressed.

Looking through your pull request, there are a few breaking changes - to OplogThread and to the oplog.timestamp format. Unfortunately, we have to be super careful when modifying the oplog.timestamp format so we don't cause mongo-connector to break when using an oplog.timestamp file from a previous version. I'm going to add this bug to my TODO list and try to come up with a solution that allows backwards compatibility, but this PR has the right idea.

Thank you for your contribution!

llvtt:
@XDestination , thanks for the pull request! I just made a very similar change in #469 that uses the replica set names as the keys now. I also needed to make sure there was an easy upgrade path from the previous file format. This issue should be fixed now.

XDestination:
Great, thank you

PR #403, opened 2016-03-04
Update links to match new owner of repo

eagleeye:


aherlihy:
Thanks for catching this! These changes are already included in the next release, which includes the elastic2.x support (which will be arriving any day now). Thank you for your contribution!

PR #412, opened 2016-03-23
Allow full class path imports with new config arg.

estobbart:
Allows an override on the module loading for a Django style import of the class.

    docManagerClassPath : 'foo.bar.baz.MyDocManager'

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -776,7 +776,7 @@ def get_config_options():
             if not isinstance(dm, dict):
                 raise errors.InvalidConfiguration(
                     "Elements of docManagers must be a dict.")
-            if 'docManager' not in dm:
+            if 'docManager' not in dm and 'docManagerClassPath' not in dm:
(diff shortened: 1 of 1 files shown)


aherlihy:
Hello, thank you for contributing!

Would you mind providing some clarification on what the use case for this is? I'm not sure what Django-style import means. Custom doc managers are already supported by using namespace packaging.

estobbart:
Sorry, should have clarified. This eliminates the need for specific directory layout. While in development you don't need all the quirky import statements in your source package's __init__.py files, or if you've deployed your app the projects can live in their own site-packages space. It also allows you to name the class whatever you want.
This is very similar to how middleware classes are imported in a Django settings file.

aherlihy:
Thank you for contributing!

estobbart:
Is there a repo for the docs where I can write up this solution?

aherlihy:
The only documentation lives in the Github Wiki, which you should be able to edit directly.

Issue #339, opened 2015-10-12
Failed during dump collection cannot recover!


snowaved:
when i sync data  from  cluter 1  to   cluster 2 ,  mongo connector runs  some time  ,it occurs some errors and exit:

2015-10-12 17:58:47,586 [CRITICAL] mongo_connector.oplog_manager:543 - Exception during collection dump
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/mongo_connector/oplog_manager.py", line 495, in do_dump
    upsert_all(dm)
  File "/usr/lib/python2.6/site-packages/mongo_connector/oplog_manager.py", line 479, in upsert_all
    dm.bulk_upsert(docs_to_dump(namespace), mapped_ns, long_ts)
  File "/usr/lib/python2.6/site-packages/mongo_connector/util.py", line 32, in wrapped
    return f(*args, **kwargs)
  File "/usr/lib/python2.6/site-packages/mongo_connector/doc_managers/mongo_doc_manager.py", line 180, in bulk_upsert
    meta_bulk_op.execute()
  File "/usr/lib64/python2.6/site-packages/pymongo/bulk.py", line 587, in execute
    return self.__bulk.execute(write_concern)
  File "/usr/lib64/python2.6/site-packages/pymongo/bulk.py", line 433, in execute
    return self.execute_command(generator, write_concern)
  File "/usr/lib64/python2.6/site-packages/pymongo/bulk.py", line 300, in execute_command
    raise BulkWriteError(full_result)
BulkWriteError: batch op errors occurred
2015-10-12 17:58:47,599 [ERROR] mongo_connector.oplog_manager:551 - OplogThread: Failed during dump collection cannot recover! Collection(Database(MongoClient([u'ip:25019', u'ip:25017']), u'local'), u'oplog.rs')
2015-10-12 17:58:47,599 [DEBUG] mongo_connector.oplog_manager:167 - OplogThread: Last entry is the one we already processed.  Up to date.  Sleeping.
2015-10-12 17:58:48,500 [ERROR] mongo_connector.connector:321 - MongoConnector: OplogThread <OplogThread(Thread-2, started 139671810971392)> unexpectedly stopped! Shutt
ing down
2015-10-12 17:58:48,501 [INFO] mongo_connector.connector:359 - MongoConnector: Stopping all OplogThreads
2015-10-12 17:58:48,501 [DEBUG] mongo_connector.oplog_manager:330 - OplogThread: exiting due to join call.
[...]

PR #431, opened 2016-04-13
log BulkWriteError in mongo_doc_manager

oxymor0n:
Log the BulkWriteError explicitly so users can debug what goes wrong during collection dump, avoiding the situation in #339


aherlihy:
Thank you for contributing! This is great.

PR #336, opened 2015-10-09
add option to skip cursor.count() to improve performance

quentins:
While the count is nice for debugging purposes, the oplog
is not optimized for the scan that count() requires so it can
take a significant amount of time. The only information we
really need is whether there are more docs to process or not.

PR #462, opened 2016-05-27
Replace count calls with cursor.next() check

jmmk:
On a large oplog (ours has 400+ million documents), calling `cursor.count()` is extremely slow. Since the count is not used outside of checking for "hast at least one document", it makes more sense to simply check if `cursor.next()` is successful.


jmmk:
I see there's another similar PR #336

jmmk:
closing in favor of #336

Issue #451, opened 2016-05-20
Process doesn't exit after fatal exception


quintstoffers:
I'm running mongo-connector in a Docker container, and it seems like the process doesn't exit when it encounters a fatal exception. Because of this the container doesn't automatically restart, meaning my MongoDB and ElasticSearch aren't synced until I manually restart.

The container does restart properly when I don't pass some required parameters (such as the MongoDB URL), so it seems like some exceptions do cause the process to exit, but others don't.

Here's the exception that occurs before the process hangs. The process doesn't print anything after this, nor does it sync anything.
```
[mongo-connector-1]2016-05-20T12:11:32.329932128Z 2016-05-20 12:11:32,314 [ERROR] mongo_connector.util:87 - Fatal Exception 
[mongo-connector-1]2016-05-20T12:11:32.329935834Z Traceback (most recent call last): 
[mongo-connector-1]2016-05-20T12:11:32.329939647Z   File "/usr/local/lib/python3.4/site-packages/urllib3/connectionpool.py", line 385, in _make_request 
[mongo-connector-1]2016-05-20T12:11:32.329943392Z     httplib_response = conn.getresponse(buffering=True) 
[mongo-connector-1]2016-05-20T12:11:32.329946700Z TypeError: getresponse() got an unexpected keyword argument 'buffering' 
[mongo-connector-1]2016-05-20T12:11:32.329950152Z  
[mongo-connector-1]2016-05-20T12:11:32.329953470Z During handling of the above exception, another exception occurred: 
[mongo-connector-1]2016-05-20T12:11:32.329956872Z  
[mongo-connector-1]2016-05-20T12:11:32.329960082Z Traceback (most recent call last): 
[mongo-connector-1]2016-05-20T12:11:32.329964337Z   File "/usr/local/lib/python3.4/site-packages/urllib3/connectionpool.py", line 387, in _make_request 
[mongo-connector-1]2016-05-20T12:11:32.329967784Z     httplib_response = conn.getresponse() 
[mongo-connector-1]2016-05-20T12:11:32.329971009Z   File "/usr/local/lib/python3.4/http/client.py", line 1227, in getresponse 
[mongo-connector-1]2016-05-20T12:11:32.329976754Z     response.begin() 
[mongo-connector-1]2016-05-20T12:11:32.329980198Z   File "/usr/local/lib/python3.4/http/client.py", line 386, in begin 
[...]

Issue #402, opened 2016-03-03
 mongo-connector not reconnecting to elasticsearch after the ReadTimeoutError occurs


kawaljeet:
When the connectivity between mongo-connector and elasticsearch is broken for a short while, mongo-connector should reconnect automatically, but it is not.

I manually stopped the elastisearch instance, to break the connectivity. I got the following messages in the mongo-connector.log file (find the log messages in the end)

After 5 minutes, I restarted the elasticsearch instance. The elastic search services are running fine and it is listening to port 9200.

But mongo-connetor didnt reconnected it (after elastic search again became available). I also tried by updating the collection, but it didnt not reconnected to elasticsearch.

Is there any configuration or the run time parameter, such that it can reconnect to the elasticsearch instance ?

I have also put the configuration file in the end
Log message

2016-03-02 22:13:28,770 [WARNING] elasticsearch:82 - PUT http://10.9.220.233:9200/people/people/56d7657ea4dc6d2b7c055908?refresh=false [status:N/A request:10.005s]
Traceback (most recent call last): 
File "/usr/local/lib/python2.7/dist-packages/elasticsearch-1.9.0-py2.7.egg/elasticsearch/connection/http_urllib3.py", line 78, in perform_request
response = self.pool.urlopen(method, url, body, retries=False, headers=self.headers, **kw)
File "/usr/local/lib/python2.7/dist-packages/urllib3-1.14-py2.7.egg/urllib3/connectionpool.py", line 608, in urlopen
_stacktrace=sys.exc_info()[2])
File "/usr/local/lib/python2.7/dist-packages/urllib3-1.14-py2.7.egg/urllib3/util/retry.py", line 224, in increment
raise six.reraise(type(error), error, _stacktrace)
File "/usr/local/lib/python2.7/dist-packages/urllib3-1.14-py2.7.egg/urllib3/connectionpool.py", line 558, in urlopen
body=body, headers=headers)
File "/usr/local/lib/python2.7/dist-packages/urllib3-1.14-py2.7.egg/urllib3/connectionpool.py", line 380, in _make_request
self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
File "/usr/local/lib/python2.7/dist-packages/urllib3-1.14-py2.7.egg/urllib3/connectionpool.py", line 308, in _raise_timeout
raise ReadTimeoutError(self, url, "Read timed out. (read timeout=%s)" % timeout_value)
[...]

Issue #383, opened 2016-01-28
mongo-connector oplog error after ReadTimeoutError


JuHwon:
Hi guys. I am using mongo-connector to fetch data from the mongodb into elasticsearch.

When my DB gets bigger and i do send requests to mongodb the mongo-connector sometimes raises the following error:

```
  2016-01-28 12:20:19,436 [WARNING] elasticsearch:82 - PUT http://localhost:9200/mongodb_meta/mongodb_meta/56a9f968a87401d779db2fe7?refresh=false [status:N/A request:10.604s]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/connection/http_urllib3.py", line 74, in perform_request
    response = self.pool.urlopen(method, url, body, retries=False, headers=self.headers, **kw)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 609, in urlopen
    _stacktrace=sys.exc_info()[2])
  File "/usr/local/lib/python2.7/dist-packages/urllib3/util/retry.py", line 222, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 559, in urlopen
    body=body, headers=headers)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 380, in _make_request
    self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 308, in _raise_timeout
    raise ReadTimeoutError(self, url, "Read timed out. (read timeout=%s)" % timeout_value)
ReadTimeoutError: HTTPConnectionPool(host=u'localhost', port=9200): Read timed out. (read timeout=10)
```

a view seconds before i was starting an aggregation in the mongodb
```
[12:20:02.134] - info: starting TenMinutesAggregation Job
```
The aggregation Job also removes some entries from the mongodb.
It seems that ElasticSearch had a small time frame where it was not responsing.

Is it normal that the mongo-connector is crashing in such a case?

And now when i try to restart the mongo-connector it tells me:
```
2016-01-28 12:20:22,920 [ERROR] mongo_connector.util:87 - Fatal Exception
[...]

PR #468, opened 2016-06-08
#383 & #402, #451 fixes for error handling and graceful process exit

pdspicer:
Addresses issues that various individuals have had with the process hanging after an error and failing to continue synchronizing after certain exceptions that should have been handled. 
The first part of this fix changes the exception_wrapper to check the class hierarchy of exceptions for wrapping.
The second part deals with the status checking of threads during the connector.py process loop, and adds a thread.is_alive() check to those threads, so that if they exit without cleaning up they can be joined and the process can exit cleanly.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -297,7 +297,8 @@ class Connector(threading.Thread):
             oplog.start()
 
             while self.can_run:
-                if not (self.shard_set[0].running and self.shard_set[0].is_alive()):
+                shard_thread = self.shard_set[0]
+                if not (shard_thread.running and shard_thread[0].is_alive()):
                     LOG.error("MongoConnector: OplogThread"
                               " %s unexpectedly stopped! Shutting down" %
                               (str(self.shard_set[0])))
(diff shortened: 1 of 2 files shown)


mongo_connector/connector.py
```diff
@@ -315,7 +315,7 @@ def run(self):
                 for shard_doc in main_conn['config']['shards'].find():
                     shard_id = shard_doc['_id']
                     if shard_id in self.shard_set:
-                        if not self.shard_set[shard_id].running:
+                        if not (self.shard_set[shard_id].running and self.shard_set[shard_id].is_alive()):
```
llvtt: Could you move the "self.shard_set[shard_id].is_alive()" to the following line, so that lines don't exceed 80 characters in length?

    if not (self.shard_set[shard_id].running and
            self.shard_set[shard_id].is_alive()):
        LOG.error(...)

or

    shard_thread = self.shard_set[shard_id]
    if not (shard_thread.running and shard_thread.is_alive()):
        LOG.error(...)

mongo_connector/util.py
```diff
@@ -32,7 +32,12 @@ def wrapped(*args, **kwargs):
                 return f(*args, **kwargs)
             except:
                 exc_type, exc_value, exc_tb = sys.exc_info()
-                new_type = mapping.get(exc_type)
+                new_type = None
+                for src_type, dest_type in mapping.iteritems():
```
llvtt: `iteritems` doesn't exist in Python 3, so iterate the keys instead:

    for src_type in mapping:
       dest_type = mapping[src_type]
       if issubclass(exc_type, src_type):
           ...

llvtt:
Thanks for the pull request! The tests are failing right now because of the `iteritems` thing, but once that's fixed (and one other minor style issue), I think this will be ready to merge.

pdspicer:
Taken care of, sorry about that, not really familiar with python & 2.x-3.x version inconsistencies.. Thanks for reviewing!

llvtt:
@pdspicer Thanks for the patch!

Issue #455, opened 2016-05-24
60% mongod-CPU Usage on IDLE


BlueRayONE:
Using MongoDB 3.2.4 an ElasticSearch 2.3.1 I'm experiencing on IDLE a CPU load of ~60% from a mongod-process.
The mongoDB-log is filled every 3 seconds with such an entry:
`2016-05-24T01:32:54.296+0200 I COMMAND  [conn598] command local.oplog.rs command: count { count: "oplog.rs", query: { ts: { $gte: Timestamp 1462970034000|1 } } } planSummary: COLLSCAN keyUpdates:0 writeConflicts:0 numYields:35407 reslen:62 locks:{ Global: { acquireCount: { r: 70816 } }, Database: { acquireCount: { r: 35408 } }, oplog: { acquireCount: { r: 35408 } } } protocol:op_query 1688ms
`

So I assume, that this is the mongo-connector, who is polling the oplog(-timestamp) from mongoDB.
There are two questions now: 

1. Why does this simple query needs so much CPU?
2. How to decrease the needed IDLE-ressources? First I'd think of decreasing the polling-rate for example to one minute. But how? I hope I didn't miss the corresponding config-option on my research.

behackett:
What version of mongo-connector  and pymongo are you using?

BlueRayONE:
I think the newest: mongo-connector 2.3 and pymongo 3.2.2

jaredkipe:
@BlueRayONE by any chance do you have some documents fail to index (e.g. because a type was different)?

BlueRayONE:
@jaredkipe I have no idea, didn't see anything. Where does this get visible? In the log? There is everything filled with the upper mentioned message. No change since my request...

behackett:
The count command is expensive because there are no indexes on the oplog (by design). During normal operation this isn't a problem, we only need to call count once when the cursor tailing the oplog is first created. Unfortunately, if the source mongod is idle for long periods of time the count command will be called often. Solving this problem is going to require some thought.

behackett:
Actually, in your case we're just querying on the ts field, which should be fast since we use the oplogReplay option. This requires more research.

behackett:
I think my original triage of this was correct. We're using Cursor.count() in get_last_oplog_timestamp. The count command isn't affected by oplogReplay (and the Cursor instance isn't using it either way). We can add ".add_option(8)" to the cursor, then do:

```
try:
    return cursor.next()
except StopIteration:
[...]

PR #470, opened 2016-06-10
Do not call count() on cursors.

llvtt:
Addresses #455 

Making a PR for travis testing.


behackett:
Seems reasonable to me. It's unfortunate that we're querying on 'ns' instead of 'ts', so oplogReplay won't help.

Issue #477, opened 2016-06-20
Update of nested properties filtered out


robertaistleitner:
I have an issue when using nested properties, it seems the ```filter_oplog_entry``` method in the oplog manager filters out valid updates with $set on nested properties.

I defined my fields like this, which works well for initial synchronization:
```
--fields _id,prop1,prop2,complexprop1.subprop1,complexprop1.subprop2
```

The problem is when I'm updating data in mongodb using $set updates with a nested property, e.g.
```
db.coll.update({"_id": ObjectId(...)}, {$set: {"complexprop1.subprop1": "value1"}})
```
it's not updated in elasticsearch due to the ```filter_oplog_entry``` method, which is not handling this case. In general this function is kind of hard to read, because it's returning something, but also modifying the passed parameter and all of this in some different ways so when reading it it's hard to understand what it actually does.

robertaistleitner:
Another little note: I already did a fix on my own for a previous version and created a pull-request as well, but it was rejected because there was another solution for it.. You can find the pull-request here: #382

ShaneHarvey:
Fixed by https://github.com/mongodb-labs/mongo-connector/commit/293e3d45d4def8920e980847e4af8a57382b097a

PR #478, opened 2016-06-20
support nested properties (fix for oplog field filtering)

robertaistleitner:
this should fix my created ticket #477, which solves issues with nested property updates.

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -371,7 +371,7 @@ class OplogThread(threading.Thread):
         self.running = False
         threading.Thread.join(self)
 
-    def _pop_excluded_fields(self, doc):
+    def _pop_excluded_fields(self, doc, update=False):
         # Remove all the fields that were passed in exclude_fields.
         for field in self._exclude_fields:
             curr_doc = doc
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -389,23 +389,31 @@ class OplogThread(threading.Thread):
                 remove_up_to.pop(end)
         return doc  # Need this to be similar to copy_included_fields.
 
+    def _contains_field(self, selector, doc, pos=0, path=[]):
+        # check if a given document contains allowed properties and return
+        # value and path for it
+        # parameter selector is a list of property names
+        # pos is internally used to know where in the selector we are
+        # path is used to be able to return a path for the new_doc for updating
+
+        if pos < len(selector):
+            for selpos in range(len(selector) - pos):
+                sel = ".".join(selector[pos:pos+selpos+1])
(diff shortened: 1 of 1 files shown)


robertaistleitner:
I have a question regarding test case test_nested_fields, which is not correct in my eyes (assertion error can be found attached). When using nested properties with elasticsearch, you need to define the fields like I did (```{'a.b.c': 2, 'a.e': 5}```), because otherwise existing subproperties will be overridden in elasticsearch. E.g. if you have ```a.c``` with a value of 10, this will be lost. Maybe the behaviour of solr is different here (I'm not using solr at all).

```
FAIL: test_nested_fields (tests.test_filter_fields.TestFilterFields)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/mongodb-labs/mongo-connector/tests/test_filter_fields.py", line 506, in test_nested_fields
    check_nested(document, fields, filtered_document)
  File "/home/travis/build/mongodb-labs/mongo-connector/tests/test_filter_fields.py", line 494, in check_nested
    self.assertEqual(filtered_result, filtered_document)
AssertionError: {'a.e': 5, 'a.b.c': 2} != {'a': {'b': {'c': 2}, 'e': 5}}
- {'a.b.c': 2, 'a.e': 5}
+ {'a': {'b': {'c': 2}, 'e': 5}}
```

letsgolesco:
Question: does this PR cover arrays? 

I've noticed that Mongo Connector misses updates to elements of arrays (whether they're strings or objects).

ShaneHarvey:
@letsgolesco this PR does not cover arrays. Right now the ``fields`` and ``exclude-fields`` options [do not support spanning arrays](https://github.com/mongodb-labs/mongo-connector/blob/2.4.1/mongo_connector/connector.py#L664-L665).

ShaneHarvey:
@robertaistleitner Your pull request looks good. Would you mind adding some tests for $set operations with include fields and exclude fields (similar [``test_nested_fields``](https://github.com/mongodb-labs/mongo-connector/blob/2.4.1/tests/test_filter_fields.py#L485) and [``test_nested_exclude_fields``](https://github.com/mongodb-labs/mongo-connector/blob/2.4.1/tests/test_filter_fields.py#L554)) so we can make sure they work properly?

letsgolesco:
Thanks for the heads up @ShaneHarvey - does that mean edits within arrays would work if the `fields` option is not used?

ShaneHarvey:
Yes, updates to arrays without using ``fields`` work properly.

ShaneHarvey:
@robertaistleitner thanks for your work on this pull request. Your changes will be merged in as part of pull request #534 which implements the remaining improvements on top of yours.

PR #479, opened 2016-06-20
Do not lose updates on docman operation failure

rectalogic:
Currently if any docman raises `OperationFailure` while processing an operation, mongo-connector just continues processing further oplog entries and updates the oplog.timestamp.

This means back end systems lose all updates that occurred during any temporary outage. We use a PostgreSQL docmanager and it is full of missing and out of date documents due to restarting postgres. mongo-connector handled the failure by just continying with the next oplog entry, and the failed entry will never be replicated.

This change breaks out of the mongo oplog cursor iteration loop if any docman fails, the cursor is then re-established and the failing oplog entry is reprocessed. The change to oplog_manager.py is almost all whitespace (indentation).

### DIFF
--- a/mongo_connector/errors.py
+++ b/mongo_connector/errors.py
@@ -26,7 +26,12 @@ class ConnectionFailed(MongoConnectorError):
 
 
 class OperationFailed(MongoConnectorError):
-    """Raised for failed commands on the destination database
+    """Raised for permanently failed commands on the destination database
+    """
+
+
+class TransientOperationFailed(MongoConnectorError):
+    """Raised for failed commands on the destination database. The command should be retried.
     """
 
 
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -315,6 +315,10 @@ class OplogThread(threading.Thread):
                             LOG.exception(
                                 "Unable to process oplog document %r"
                                 % entry)
+                        except errors.TransientOperationFailed:
+                            LOG.exception(
+                                "Transient failure while processing oplog document %r"
+                                % entry)
                             cursor.close()
                             break
                         except errors.ConnectionFailed:
@@ -879,7 +883,7 @@ class OplogThread(threading.Thread):
                         remov_inc += 1
                         LOG.debug(
                             "OplogThread: Rollback, removed %r " % doc)
-                    except errors.OperationFailed:
+                    except (errors.OperationFailed, errors.TransientOperationFailed):
                         LOG.warning(
                             "Could not delete document during rollback: %r "
                             "This can happen if this document was already "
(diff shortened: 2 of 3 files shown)


rectalogic:
Use this url to view the PR ignoring whitespace https://github.com/mongodb-labs/mongo-connector/pull/479/files?w=1

llvtt:
Thanks for the pull request! In the case where the remote system has just gone down temporarily, I think it makes a lot of sense to retry the operation. There are some cases, though, where the operation is simply impossible, and the best we can do is log the fact that it can't work and continue on. For example, the Mongo doc manager will raise OperationFailed if we try to call `renameCollection` and the target collection already exists (this could happen if replaying the oplog from an earlier point).

Perhaps the best way to proceed is to make this retry behavior optional. If you agree, would you mind adding an option for it? Perhaps something like `retryOnError`?

rectalogic:
What if we define a new `TransientOperationFailed` error? Continue to treat `OperationFailed` as unrecoverable, but a DocManager can instead raise `TransientOperationFailed` if it knows the error is recoverable and should be retried.

rectalogic:
@llvtt does the above change look like a reasonable alternative to adding an option?

llvtt:
The more I think about this, the more it seems that the correct way forward is to have the doc managers themselves do the retrying, rather than putting this into mongo-connector itself.

I think having the specialized Exception type will depend too much on docman authors updating their doc managers to take advantage of this new Exception. Additionally, some doc managers/underlying drivers don't provide specific enough error messaging to determine if an operation can be retried (e.g. pysolr raises `SolrError` as its only Exception type for all errors, regardless of cause). The result will be that whether mongo-connector does the Right Thing with regards to retrying certain operations will be hard to determine. We may as well move this decision to the doc managers themselves. Users can consult the documentation of the individual doc managers to investigate retry behavior.

The tunable "retryOnError" flag I was proposing earlier isn't a great idea, either, since it's too coarse-grained. Some OperationFailures can probably be retried, others cannot. Stopping/starting mongo-connector to get past certain errors and not others by turning on/off this flag will be pretty annoying and brittle.

So, I suggest making a pull request against the PostgreSQL doc manager to retry any operations that it can. It's also worth noting that doc managers themselves can take in configuration options from mongo-connector's own config file, so if you want to make it tunable at the doc manager level, you can: https://github.com/mongodb-labs/mongo-connector/wiki/Configuration-Options#docmanagersindexargs

rectalogic:
OK. The postgres connector is my own so I'll put reconnect logic in it.
https://github.com/cureatr/postgresql-doc-manager

llvtt:
Great! Thanks for the discussion. Please let me know if there's anything you need to support this.

PR #487, opened 2016-07-06
Continue on BSON parsing errors if continue_on_error is true.

sha1sum:
BSON parsing errors should not cause critical-level errors if `continue_on_error` is `true`. This PR allows records to continue to be processed on a dump in cases where these parsing errors occur.

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -504,6 +504,8 @@ class OplogThread(threading.Thread):
             database, coll = namespace.split('.', 1)
             last_id = None
             attempts = 0
+            to_skip = 0
+            cursor_errors = 0
 
             # Loop to handle possible AutoReconnect
             while attempts < 60:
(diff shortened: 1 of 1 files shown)


behackett:
What BSON error are you trying to work around?

sha1sum:
@behackett There were a few "Date out of range" `InvalidBSON` errors when attempting to loop through the documents in the cursor.

behackett:
Ah, good old python datetime.datetime, which only accepts years between 1 and 9999. Are you sure you have PyMongo's C extensions installed? Datetime's C API accepts much larger values for year. You can check like this:

python -c "import pymongo; print(pymongo.has_c())"

sha1sum:
That's a `False`. I'll try to see how to get them installed. I'm a golang dev, so Python is not my speciality, and this is an AWS EC2 instance, but I'll see what I can do!

behackett:
There are no pre-compiled binaries for Linux on pypi (it's a long sad story about python packaging that will be fixed soon), so make sure you follow these instructions before reinstalling pymongo:

http://api.mongodb.com/python/current/installation.html#dependencies-for-installing-c-extensions-on-unix

sha1sum:
Opting for a log warning via #505 -- this PR is no longer necessary.

Issue #488, opened 2016-07-12
Extend namespaces.mapping to support wildcards


weixili:
If we have features that supporting wildcards in namespaces.include/exclude ( #285 #480 ), it would be good that also extending namespaces.mapping to support similar behaviors. In our case, we may have different index for different collections, but it would be nice that we dont need to specify all collections one by one. 

For example, we would like it supports mapping like: `foo.*` to `bar_*.someting`, so if we have two collections `foo.a` and `foo.b` will map to `bar_a.something` and `bar_b.something`.

llvtt:
I think this makes sense. So, the destination collection/type will have a name that is the value from the namespace mapping, with the wildcard character (`*`) replaced by whatever the wildcard character matched in the source. I believe this matches your examples above.

ShaneHarvey:
mongo-connector now has support for excluding namespaces as well as using a wildcard in namespaces.include/exclude/mapping. [See the wiki for details](https://github.com/mongodb-labs/mongo-connector/wiki/Configuration-Options#configure-namespaces). Thanks @weixili!

Issue #480, opened 2016-06-22
Support wildcard characters in namespace include/exclude options


PeterBackman:
The default behaviour of mongo-connector is to automatically synchronize all collections in all databases with elasticsearch (or other systems). This is really powerful but in our system we have different databases in MongoDB with different purposes. Due to this we only want to synchronize a subset of databases with elasticsearch. This is possible by specifying namespace.include configuration but this kind of micro management of collections is cumbersome since our system will add/remove collections based on user input.

It would be really great if one could limit the amount of databases to synchronize without having to specify every collection to synchronize.

llvtt:
Perhaps `namespaces.include` and `namespaces.exclude` could accept wildcard characters like `*` that match anything. For example, one could specify `{"namespaces": {"include": ["foo.*"]}}` to mean "synchronize all collections in the `foo` database". Would that fit your use case? This is something that has been requested before in #285.

PeterBackman:
Sorry didn't see #285. This solution with wildcards and an exclude option would would work fine for us for our new development.

Would namespace.mapping still be hardcoded names or could that be extended as well? Either way this would simplify a lot for us.

llvtt:
The scope of this feature is just using wildcards in namespace include/exclude. Extending namespace mappings to map names like `foo.*` to `bar.*` is a separate issue. Is this something you're interested in as well? If so, please feel free to open another issue and I'll take a look.

weixili:
Hi, I submitted a new issue about the namespaces.mapping #488. Please have a look at it.

weixili:
I thought to implement this feature dynamically. For example, if there are `foo.a`, `foo.b` and `foo.c` in our db and we specify to include `foo.*` in the mongo-connector configuration, now we have a new collection created called `foo.d`, we dont need to restart mongo-connector but the change can be detected automatically. But when I had a look at the code handling cursors, it seems like it needs a real big change in the code. 

It seems like the current code is more like a static way, am I right?
It would be much easier to support wildcard in a static way. If we have a new collection, it has to restart the mongo-connector to let it knows the change. Would that be acceptable?

llvtt:
@weixili, I believe this feature would include picking up automatically a new collection that was created after the connector has been started. No one has started work on this feature yet, but I don't think it will be too tricky.

letsgolesco:
Question: are you guys planning on supporting wild cards for collections only, or for databases as well?

[...]

PR #495, opened 2016-07-22
Support wildcard feature in namespaces.include/exclude/mapping

weixili:
Support configuration for example:

"namespaces": {
		"include": [
		  "foo.\*"
		],
		"exclude": [
	      "foo.jobs"
	    ],
	    "mapping": {
	      "foo.\*": "foo000_\*.documents"
	    }
	}

More details please see related issues:
https://github.com/mongodb-labs/mongo-connector/issues/488
https://github.com/mongodb-labs/mongo-connector/issues/480


weixili:
Hi @llvtt,

This is my commit for the new features #488 #480 . Please have a review on this. It's my first time to make a pull request so I am not sure if my way is correct or not. If anything is wrong, please let me know. 

I tested with our mongo-db and elasticsearch and it met our requirements. But I have problems to run your tests locally, so did not test/change it. Hopefully it works.

weixili:
Finally, the tests are all passed. Please have a review on this change if you have time. Thanks!

llvtt:
Hi, glad you got the tests working! I'll take a look at this as soon as I can. Thanks for your hard work on this.

mongo_connector/command_helper.py
```diff
@@ -16,59 +16,29 @@
...
-                       dest))
-            else:
-                targets.add(dest)
-
-        self.db_mapping = {}
-        for src, dest in db_pairs:
-            arr = self.db_mapping.get(src, [])
-            arr.append(dest)
-            self.db_mapping[src] = arr
+    def __init__(self, dest_mapping_stru=DestMapping([],[],{})):
```
llvtt: Don't use mutable objects as default values for keyword arguments. See http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments. Here, it seems that "dest_mapping_stru" is actually a required argument, because I don't see any places where we don't pass an instance of DestMapping, so it doesn't need a default value.

mongo_connector/connector.py
```diff
@@ -25,13 +25,15 @@
 import sys
 import threading
 import time
+import collections
```
llvtt: This import is unused.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+from mongo_connector import errors
+
+
+LOG = logging.getLogger(__name__)
+
+"""New structure to handle both plain and wildcard mapping namespaces dynamically.
+"""
+
+class DestMapping():
+    def __init__(self, namespace_set=[], ex_namespace_set=[], inpt={}):
```
llvtt: Same issue here with the default values for keyword arguments.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+        # because the mappings are not duplicated, so the values should also be unique
+        self.reverse_plain = {}
+        # a dict containing plain db mappings, db -> an array of mapped db
+        self.plain_db = {}
+        
+        # the input namespace_set and ex_namespace_set could contain wildcard
+        self.namespace_set = namespace_set
+        self.ex_namespace_set = ex_namespace_set
+        
+        self.lock = threading.Lock()
```
llvtt: Can you please explain why locking is necessary? It seems that the user should specify namespaces and the mapping before running the connector, then this object should remain immutable after that.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+        
+        # initialize
+        for ns in namespace_set:
+            if ns not in inpt.keys():
+                inpt[ns] = ns
```
llvtt: This would be a good use case for [setdefault](https://docs.python.org/3/library/stdtypes.html?highlight=setdefault#dict.setdefault).

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+            
+    def match(self, src, dst_pattern):
+        """If source string src matches dst, return the matchobject"""
+        d_arr = re.split('(\*)',dst_pattern)
+        r_arr = []
+        for x in d_arr:
+            if x != "*":
+                r_arr.append(re.escape(x))
+            else:
+                r_arr.append("([^*]+)")
```
llvtt: I think this can be easier. The patterns `\A` and `\Z` can be used to represent the beginning/end of a string in Python, respectively. So, I think `reg_pattern` could just be `r'\A' + dst_pattern.replace('*', '.*') + r'\Z'`.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+            return None
+        pattern_arr = re.split('(\*)', map_pattern)
+        map_arr = []
+        map_index = 1
+        for x in pattern_arr:
+            if x != "*":
+                map_arr.append(x)
+            else:
+                map_arr.append(src_match.group(map_index))
+                map_index += 1
```
llvtt: I think this can be made easier, too. If there's only one wildcard in the map_pattern, then we can just substitute the group from src_match into where the "*" is in map_pattern like this:

    wildcard_matched = src_match.group(1)
    return map_pattern.replace("*", wildcard_matched)

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+                m = self.match(plain_src, x)
+                if m:
+                    return True
+        
+        return False
+    
+    def replace(self, src_match, map_pattern):
+        """Given the matchobject src_match, replace corresponding '*' in map_pattern with src_match."""
+        num_match = src_match.lastindex
+        if num_match != map_pattern.count("*"):
```
llvtt: I'm confused about what this is validating. It looks like we're comparing the index of the match from the source to the number of `*` characters in map_pattern?

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+        else:
+            # search in wildcard mappings
+            # if matched, get a replaced mapped namespace
+            # and add to the plain mappings            
+            for k, v in self.wildcard.items():
+                m_res = self.match(plain_src_ns, k)
+                if m_res:
+                    res = self.replace(m_res, v)
+                    if res is not None:
+                        if res not in self.reverse_plain:
```
llvtt: It seems like a lot of this code is doing what `self.set` does. Perhaps use that method here?

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def set(self, key, value):
+        """Set the DestMapping instance to corresponding dictionary"""
+        self.lock.acquire()
+        db = key.split(".")[0]
+        if "*" in db:
+            raise errors.InvalidConfiguration("Wildcard is not allowed in db name")
+            
+        if "*" in key:
+            if key.count("*") != value.count("*"):
```
llvtt: I think we should limit the number of `*` characters in any pattern to 1. Any more than that, and this feature gets very complicated. Does your use case require being able to match more than one wildcard character in a namespace?

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+                raise errors.InvalidConfiguration("The number of wildcards in the mapping should be equal")
+            self.wildcard[key] = value
+        else:
+            if value not in self.reverse_plain:
+                self.plain[key] = value
+                self.reverse_plain[value] = key
+                arr = self.plain_db.get(db, [])
+                arr.append(value.split(".")[0])
+                self.plain_db[db] = list(set(arr))
```
llvtt: Perhaps self.plain_db should be a mapping of strings -> set objects. When adding something to self.plain_db[db], let's just call `self.plain_db[db].add(<something>)`

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.match_set(plain_src_ns, self.ex_namespace_set):
+            return None       
+        # if no namespace_set, no renaming
+        if not self.namespace_set:
+            return plain_src_ns
+        # if plain_src_ns does not match namespace_set, ignore
+        elif not self.match_set(plain_src_ns, self.namespace_set):
+            return None
+        else:
+            return self.get(plain_src_ns, plain_src_ns)
```
llvtt: Can we combine these last two cases by just returning `self.get(plain_src_ns)`? If get() doesn't find plain_src_ns in self.namespace_set, it should return `None` anyway.

mongo_connector/oplog_manager.py
```diff
@@ -74,11 +75,14 @@ def __init__(self, primary_client, doc_managers,
...
+        # The set of namespaces not to process from the mongo cluster.
+        self.ex_namespace_set = kwargs.get('ex_ns_set', [])
+
         # The set of gridfs namespaces to process from the mongo cluster
         self.gridfs_set = kwargs.get('gridfs_set', [])
 
-        # The dict of source namespaces to destination namespaces
-        self.dest_mapping = kwargs.get('dest_mapping', {})
+        # The structure handling dest_mappings dynamically
+        self.dest_mapping = dest_mapping_stru
```
llvtt: I don't think we need to change the constructor signature. We can still pull "dest_mapping" out of kwargs. I don't think we need to change the parameter name either.

mongo_connector/oplog_manager.py
```diff
@@ -183,12 +196,42 @@ def oplog_ns_set(self):
...
             self._oplog_ns_set.extend(self.gridfs_files_set)
-            self._oplog_ns_set.extend(set(
-                ns.split('.', 1)[0] + '.$cmd' for ns in self.namespace_set))
             self._oplog_ns_set.append("admin.$cmd")
 
+    @property
+    def oplog_ex_ns_set(self):
+        try:
+            return self._oplog_ex_ns_set
+        except AttributeError:
```
llvtt: This shouldn't ever happen, since we call `self.oplog_ex_ns_set = ...` in the constructor, which sets `self._oplog_ex_ns_set`.

mongo_connector/oplog_manager.py
```diff
@@ -479,26 +531,84 @@ def dump_collection(self):
...
-                coll_list = retry_until_ok(
-                    self.primary_client[database].collection_names)
-                for coll in coll_list:
-                    # ignore system collections
-                    if coll.startswith("system."):
-                        continue
-                    # ignore gridfs collections
-                    if coll.endswith(".files") or coll.endswith(".chunks"):
+
+        def get_dump_set():
```
llvtt: Since this is only ever called from one place and doesn't take any arguments, it doesn't need to be a function. We can just modify `dump_set` directly.

llvtt:
Hi @weixili,

I've started reviewing your pull request. I have a few comments overall:

1. I think we should allow only one `*` character in a namespace for matching. I think this will meet 99.9% of the use cases out there, and it simplifies the code.

2. We should only allow a namespace set *or* an "exclude" namespace set, but not both. I think this also should simplify some things.

3. As much as possible, the code in this repository tries to adhere to [PEP8](https://www.python.org/dev/peps/pep-0008/) style guidelines. You don't have to read through the whole style guidelines, because there are handy command-line tools like [pep8](https://pypi.python.org/pypi/pep8) that can do the checking for you. The pep8 tool may also catch some problems in mongo-connector itself, but don't worry too much about these. My concern is that this PR follow the same style as the rest of the project.

weixili:
Hi @llvtt,

Thanks for your review! I will change the code soon accordingly. Here are some comments about the general issues. Please confirm we aggree on the same level. 


> 1.  I think we should allow only one * character in a namespace for matching. I think this will meet 99.9% of the use cases out there, and it simplifies the code.
> 2. We should only allow a namespace set or an "exclude" namespace set, but not both. I think this also should simplify some things.

Both are okay for us. Let me list some allowed and not-allowed wildcard examples to see if we agree:

For include/exclude:
- Only one * character appears in the collection name and no other characters:

e.g. allow: 
`"include": [ "foo.*" ]`

but not allow:
` "include": [ "foo_*.bar" ] `

or 

`"include": ["foo.*abc"] `

or 

`"include": ["foo.*abc*cde"]`

- Only allow "include" or "exclude", not both

e.g. allow: 
```
"namespaces": {
    "include": [ "db1.*", "db2.*" ]
  }
```

or 

```
"namespaces": {
    "exclude": [ "db1.*", "db2.*" ]
  }
```

but not allow:
```
"namespaces": {
    "include": [ "db1.*", "db2.*" ],
    "exclude": [ "db1.col1"]
  }
```
so if you want to include all collections in db1 except col1, the user needs to list all allowed collections in db1 in the include. For us, we dont have this kind of requirement yet.

For mapping:
- The key follows the same rule as the include/exclude, and the mapped value can only have one * character in either db name or collection name and can include other plain characters:

e.g. allow:
```
"mapping": {
        "db1.*" : "newdb1_*.newcol",
        "db2.*" : "newdb2.*"
    }
```

but not allow:
```
"mapping": {
        "db1.*" : "newdb1_*x*.newcol",
        "db2.*" : "newdb2_*.new_*"
    }
```



>  As much as possible, the code in this repository tries to adhere to PEP8 style guidelines. You don't have to read through the whole style guidelines, because there are handy command-line tools like pep8 that can do the checking for you. The pep8 tool may also catch some problems in mongo-connector itself, but don't worry too much about these. My concern is that this PR follow the same style as the rest of the project.

Yes, I will check my style with the tool.

> I see you added a DestMappingTest, which tests the DestMapping class. In addition to this, it would be nice to have a test case that constructs an OplogThread that uses a DestMapping, so that we can test the mapping of namespaces with wildcards in action. You can follow the TestOplogManager class as a template TestCase for this.

No problem. I will create a new test file called like test_oplog_manager_wildcard.py with class called like TestOplogManagerWildcard. Is that okay?


Thank you!

mongo_connector/command_helper.py
weixili: Because in the test_commands of TestSolrDocManager class, there is one case that initiate CommandHelper with no argument and the old __init__ function in CommandHelper has default values as

`def __init__(self, namespace_set=[], dest_mapping={})`

so I changed accordingly, but you are right, this is not necessary and in that testcase, actually we can init a DestMapping object with empty values and then pass to the CommandHelper. I will remove this default value. Also in the DestMapping init.

mongo_connector/dest_mapping.py
weixili: The DestMapping object will be changed when there is a new wildcard mapping used by get function. And this DestMapping will be passed to create OplogThread and could be multi-thread.  I never used this Lock feature before, please correct me if anything is wrong.

mongo_connector/dest_mapping.py
weixili: This is based on the assumption that there could be multiple * characters in the mapping. So I was trying to check if the number of * characters in the mapping's key and value is equal. For example, "foo.*": "fxy_*_z_*.bar" is not correct. But if we limit the number of * character to be only one, then we dont need this check.

mongo_connector/dest_mapping.py
weixili: No, we can add this limitation then the code is much simpler.

mongo_connector/oplog_manager.py
weixili: Sorry, I dont get your point. The old dest_mapping will never be used and I passed the new DestMapping object to the OplogThread init function. I used the original variable name self.dest_mapping just because I can avoid to change all the occurrences of self.dest_mapping.get() function with new names. If you think this makes people confused, then I change all the self.dest_mapping to self.dest_mapping_stru, but do you think we need to keep the original self.dest_mapping even it is not used anymore?

llvtt:
1. What I mean by "allow only one * character in a namespace for matching" is that there should only be one `*` anywhere in a namespace. These, for example, would be invalid:

   - `foo*.bar*`
   - `foo.b*ar*`

   These are fine:

   - `foo_*.bar`
   - `foo.ba*r`

   Having only one `*` makes creating the regex pattern easy, and doing the namespace mapping when there's only one `*` is much more straightforward than for an arbitrary number of `*` characters (just replace the `*` with whatever was matched by the other `*`).

2. I think we agree. If this is requested later, we can always add it, but for now, let's avoid the additional complexity.

3. Thanks!

4. That test name sounds fine to me. Let me know if you need help running the tests. This [wiki page](https://github.com/mongodb-labs/mongo-connector/wiki/Running-the-Tests) may also come in handy.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,186 @@
...
+    
+    def map_namespace(self, plain_src_ns):
+        """Applies the plain source namespace mapping to a "db.collection" string.
+        The input parameter ns is plain text."""
+        # if plain_src_ns matches ex_namespace_set, ignore
+        if self.match_set(plain_src_ns, self.ex_namespace_set):
+            return None       
+        # if no namespace_set, no renaming
+        if not self.namespace_set:
+            return plain_src_ns
```
weixili: Hi @llvtt ,

I am a bit confused about this case. In the current no wildcard version, the mapping must be used together and both options must include the same number of names. It is valid when the user is using the command line parameters. But the code did not check the configuration file case, does that mean it is allowed to have no include but mapping in the config file?  If not allowed, when we implement the wildcard case, do we need to follow the same rule?

mongo_connector/oplog_manager.py
weixili: Hi @llvtt ,

We did not call self.oplog_ex_ns_set - ... in the constructor, but in the get_last_oplog_timestamp() and get_oplog_cursor(). In the constructor, we call self.ex_namespace_set = ....

I just followed the same way as the code handles with oplog_ns_set. I feel complicated about the part of code as well.

What is your opinion? Do I need to remove all try except for oplog_ns_set, gridfs_files_set and  oplog_ex_ns_set?

llvtt: Ooops - I meant to say that we call [`self.ex_namespace_set = ...`](https://github.com/mongodb-labs/mongo-connector/pull/495/files/19ab0ed1658d184061f3312994cb811a04c745a9#diff-9e5fc676a1f3b77e19871877266b4cf0R79) in the constructor, which calls `update_oplog_ex_ns_set`, which in turn sets self._oplog_ex_ns_set.

I don't know why the other getter methods are using try/except AttributeError, since those seem similarly impossible. I'd say remove the try/except here. I'll run the tests again to determine if the try/except cases are necessary in the other getters, though I can't see why they would be.

weixili:
Hi @llvtt , I have uploaded a new patch. Please have a review. Thank you!

weixili:
Hi, the new merged commits in master branch made my change not compatible especially the changes in oplog_manager.py. I have to update my code and may take some time.

mongo_connector/oplog_manager.py
```diff
@@ -459,6 +505,19 @@ def get_oplog_cursor(self, timestamp=None):
...
+            query['ns'] = {'$in': self.oplog_ns_set}
+
+        if self.oplog_ex_ns_set and not self.oplog_ns_set:
+            query['ns'] = {'$nin': self.oplog_ex_ns_set}
+
+        if self.oplog_ns_set and self.oplog_ex_ns_set:
+            query = {'$and': [{'ns': {'$nin': self.oplog_ex_ns_set}},
+                              {'ns': {'$in': self.oplog_ns_set}}]}
+
+        LOG.debug("Cursor query: %s" % query)
```
ShaneHarvey: The cursor returned from get_oplog_cursor no longer filters by namespace. The filtering is instead done in the run method here: https://github.com/mongodb-labs/mongo-connector/pull/495/files#diff-9e5fc676a1f3b77e19871877266b4cf0R314

If you remove these lines and add the equivalent filtering logic using `self.oplog_ns_set` and `self.oplog_ex_ns_set` up there the tests should pass again.

mongo_connector/oplog_manager.py
```diff
@@ -644,10 +730,23 @@ def do_dump(dm, error_queue):
...
+
```
ShaneHarvey: `_get_oplog_timestamp` no longer filters by namespace. I think you can simply remove these lines. Then after addressing my other comment the tests should pass again.

mongo_connector/oplog_manager.py
weixili: I am considering moving this part of logic inside the new DestMapping class. Otherwise, it needs to do regex matching which already did in that class again. And the performance will be better since inside that class there is kind of caching of the wildcard patterns.

weixili: I found I dont need to write the match logic again then I just as you said add some if statement in the run() is enough.

weixili:
Hi, this PR has been updated with no conflict. Please have a review. Thank you!

p-backman-ericsson:
Ping @llvtt and @ShaneHarvey . Could you please look at the revised patch?

mongo_connector/connector.py
```diff
@@ -704,19 +713,53 @@ def apply_namespaces(option, cli_values):
...
+        ex_ns_set = option.value['exclude']
+        if len(ex_ns_set) != len(set(ex_ns_set)):
+            raise errors.InvalidConfiguration(
+                "Exclude namespace set should not contain any duplicates.")
+
+        # not allow to exist both 'include' and 'exclude'
+        if ns_set and ex_ns_set:
+            raise errors.InvalidConfiguration(
+                "Include and exclude namespace sets"
+                " are not allowed to exist both.")
```
ShaneHarvey: Change to "Cannot use both namespace 'include' (--namespace-set) and 'exclude' (--exclude-namespace-set)."

mongo_connector/connector.py
```diff
@@ -704,19 +713,53 @@ def apply_namespaces(option, cli_values):
...
         dest_mapping = option.value['mapping']
         if len(dest_mapping) != len(set(dest_mapping.values())):
             raise errors.InvalidConfiguration(
                 "Destination namespaces set should not"
                 " contain any duplicates.")
 
+        for key, value in dest_mapping.items():
+            if key.count("*") > 1 or value.count("*") > 1:
+                raise errors.InvalidConfiguration(
+                    "Only one * character is allowed in the namespaces!")
```
ShaneHarvey: Change to "The namespace mapping source and destination cannot contain more than one '*' character.

mongo_connector/connector.py
```diff
@@ -704,19 +713,53 @@ def apply_namespaces(option, cli_values):
...
+            if key.count("*") != value.count("*"):
+                raise errors.InvalidConfiguration(
+                    "The number of * character should be equal in the mapping's key and value!")
```
ShaneHarvey: Change to "The namespace mapping source and destination must contain the same number of '*' characters."

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+        """Given the matchobject src_match,
+        replace corresponding '*' in map_pattern with src_match."""
+        wildcard_matched = src_match.group(1)
+        return map_pattern.replace("*", wildcard_matched)
+
+    def get(self, plain_src_ns, defaultval=""):
+        """Given a plain source namespace, return a mapped namespace if matched.
+        If no match and given defaultval, return it as a default value.
+        """
+        self.lock.acquire()
```
ShaneHarvey: Acquire the lock using a with statement to simplify this code. See, https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+        # search in plain mappings first
+        if plain_src_ns in self.plain.keys():
+            self.lock.release()
+            return self.plain[plain_src_ns]
+        else:
```
ShaneHarvey: I would rewrite this to be more idiomatic:
```python
with self.lock
    try:
        return self.plain[plain_src_ns]
    except KeyError:
        # search in wildcard...
```

mongo_connector/dest_mapping.py
ShaneHarvey: Can you move this check (``if res not in self.reverse_plain:``) into the ``set_plain`` method? That would simplify this code a bit. 

Can you also add more information to the ``InvalidConfiguration`` error? Some useful info would be the actual values of the namespaces that clashed. Since this will halt mongo-connector we want to be as descriptive as possible.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+            self.lock.release()
+            return defaultval
+        else:
+            self.lock.release()
+            LOG.warn("Failed to find matched mapping for %s." % plain_src_ns)
+            return None
+
+    def set(self, key, value):
+        """Set the DestMapping instance to corresponding dictionary"""
+        self.lock.acquire()
```
ShaneHarvey: Acquire the lock using a with statement to simplify this code. See, https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if "*" in key:
+            self.wildcard[key] = value
+        else:
+            if value not in self.reverse_plain:
```
ShaneHarvey: Same comment as above: move this check down into ``set_plain``.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+        self.lock.release()
+
+    def get_key(self, plain_mapped_ns):
+        """Given a plain mapped namespace,
+         return a source namespace if matched.
+        Only need to consider plain mappings
+         since this is only called in rollback
+         and all the rollback namespaces in the
+         target system should already been
+         put in the reverse_plain dictionary.
```
ShaneHarvey: Fix format.

mongo_connector/dest_mapping.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def map_namespace(self, plain_src_ns):
+        """Applies the plain source namespace mapping to a "db.collection" string.
+        The input parameter ns is plain text."""
+        # if plain_src_ns matches ex_namespace_set, ignore
+        if self.match_set(plain_src_ns, self.ex_namespace_set):
+            return None
+        # if no namespace_set, no renaming
+        if not self.namespace_set:
+            return plain_src_ns
```
ShaneHarvey: I don't think this correctly handles the case where ``namespaces.mapping`` is defined and ``namespaces.inlcude`` is not? For example a config file with,
```json
{
    "namespace": {
        "mapping": {"foo.*": "bar.*"}
    }
}
```

``map_namespace('foo.bob')`` would return 'foo.bob', when the correct mapping is 'bar.bob'.

I think the correct behavior here should be:
```python
if not self.namespace_set:
    return self.get(plain_src_ns, defaultval=plain_src_ns)
else:
    return self.get(plain_src_ns)
```

weixili:
Hi @ShaneHarvey , thanks for the review. New patch has been updated.

ShaneHarvey:
@weixili I went ahead and merged your changes into master (see 40c535ae04c1a142f48ad5f3f14a50cb4509fb54). We're excited to have this in the next release, thanks for all your hard work on this feature!

a1u:
Can you please suggest, I do not see this in 2.4.1 release. Can you please tell me when you plan the next release, so, I would be able to use wildcard feature?
Thanks.

Issue #334, opened 2015-10-09
mongo-connector stop working if a document contains invalid data  when dumping collections


Jinshui:
While dumping a collection from mongodb to elasticsearch, the mongo-connector exit because one of the document has an invalid date value. The following is the error trace:

>2015-10-09 15:46:35,840 [CRITICAL] mongo_connector.oplog_manager:543 - Exception during collection dump
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/oplog_manager.py", line 495, in do_dump
    upsert_all(dm)
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/oplog_manager.py", line 479, in upsert_all
    dm.bulk_upsert(docs_to_dump(namespace), mapped_ns, long_ts)
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/util.py", line 32, in wrapped
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/doc_managers/elastic_doc_manager.py", line 190, in bulk_upsert
    for ok, resp in responses:
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/helpers/__init__.py", line 110, in streaming_bulk
    for action, data in chunk:
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/doc_managers/elastic_doc_manager.py", line 156, in docs_to_upsert
    for doc in docs:
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/oplog_manager.py", line 442, in docs_to_dump
    for doc in cursor:
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 1090, in next
    if len(self.__data) or self._refresh():
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 1039, in _refresh
    limit, self.__id))
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 958, in __send_message
    self.__compile_re)
  File "/usr/local/lib/python2.7/dist-packages/pymongo/helpers.py", line 121, in _unpack_response
    compile_re)
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 537, in decode_all
    tz_aware, uuid_subtype, compile_re))
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 331, in _elements_to_dict
[...]

Issue #279, opened 2015-07-08
mongo-connector stuck on fatal error


joe-mojo:
I have 2 replicaset in 2 different environnment (dev&prod) with no issue. They run mongo-connector 2.0.3 with pyMongo 2.8. Prod is running with python 2.6, dev with 2.7.

I tried recently to set up a 3thd env with dev data. I ran into fatal issue. This occured with pyMongo 2.0.3, 2.10 and pyMongo 2.8 and 2.8.1.

When starting mongo-connector service, after a constant number of 6000 doc inserted into ElasticSearch. The 6001nth is raising: 
```
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/util.py", line 85, in wrapped
    func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mongo_connector/oplog_manager.py", line 185, in run
    for n, entry in enumerate(cursor):
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 1090, in next
    if len(self.__data) or self._refresh():
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 1039, in _refresh
    limit, self.__id))
  File "/usr/local/lib/python2.7/dist-packages/pymongo/cursor.py", line 958, in __send_message
    self.__compile_re)
  File "/usr/local/lib/python2.7/dist-packages/pymongo/helpers.py", line 121, in _unpack_response
    compile_re)
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 537, in decode_all
    tz_aware, uuid_subtype, compile_re))
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 331, in _elements_to_dict
    data, position, as_class, tz_aware, uuid_subtype, compile_re)
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 320, in _element_to_dict
    data, position, as_class, tz_aware, uuid_subtype, compile_re)
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 159, in _get_object
    encoded, as_class, tz_aware, uuid_subtype, compile_re)
  File "/usr/local/lib/python2.7/dist-packages/bson/__init__.py", line 331, in _elements_to_dict
[...]

PR #505, opened 2016-08-13
Log a warning regarding missing pymongo C extensions.

sha1sum:
For those that are new to Python or are not familiar with PyMongo, `InvalidBSON` errors regarding dates can be confusing (I was one of those people). If the `pymongo` package was installed without C extensions, there will now be a warning log entry notifying users of possible "Date out of range" errors.

This is a replacement for the unnecessary #487 which will be closed.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -65,6 +65,10 @@ class Connector(threading.Thread):
             LOG.warning('No doc managers specified, using simulator.')
             self.doc_managers = (simulator.DocManager(),)
 
+        # Warning when pymongo does not have C extensions regarding possible "Date out of range" errors
+        if not pymongo.has_c():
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
I think this is a great idea considering all (#279, #334, #526) the issues about the `date value out of range` errors.

ShaneHarvey:
Thanks for the pull!

shenberg:
Umm, so, in platforms without C extension support (e.g. PyPy) there is no option to follow oplogs?

uriinf:
Note that it will not save you (not anymore?) from ```InvalidBSON: date value out of range``` (at least with versions pymongo==3.8 and mongo-connector==3.1.1). Check the [issue](https://github.com/yougov/mongo-connector/issues/279#issuecomment-497228765)

PR #534, opened 2016-09-20
Properly filter dot notation fields (including arrays) in updates

### LINKED ISSUE #477: Update of nested properties filtered out

### LINKED ISSUE #520: Feature Request: `fields` option should propagate edits to array elements
I'm running Mongo Connector to replicate data from a MongoDB instance to an Elasticsearch instance. 

My config includes a `fields` option because I only want a few of my fields to be propagated, including a few array fields.

Some are arrays of strings, others are arrays of objects (including [geojson](http://geojson.org/) objects).

I've noticed that `$addToSet` and `$pull` updates in MongoDB result in the whole array being updated in Elasticsearch, which is fine.

Some important updates aren't replicated in my setup however, including:
* Overwrites of array elements (e.g. `{ $set: { "array.0": "newValue" } }`)
* Edits to objects within arrays (e.g. `{ $set: { "array.$.field": "newValue" } }`)

My feature request is for updates to array elements to be propagated by Mongo Connector when using the `fields` option.

ShaneHarvey:
Properly filter dot notation fields (including arrays) in updates. Fields and exclude fields still do not support *spanning* arrays but referencing array fields themselves is now supported.
Includes tests for filtering updates with dot notation fields and arrays.

Fixes #477 (Update of nested properties filtered out).
Resolves #520 (Feature Request: `fields` option should propagate edits to array elements). Updates such as `{ $set: { "array.0": "newValue" } }` and `{ $set: { "nested.array.0": "newValue" } }` are now included or excluded correctly.

This pull request builds on Robert Aistleitner's fix for the same issue submitted here #478. Thanks @robertaistleitner!

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -431,9 +431,9 @@ class OplogThread(threading.Thread):
         try:
             for key in path:
                 doc = doc[key]
+            return [(path, doc)]
         except (KeyError, TypeError):
             return []
-        return [(path, doc)]
 
     @classmethod
     def _find_update_fields(cls, field, doc):
@@ -479,8 +479,7 @@ class OplogThread(threading.Thread):
         find_fields = self._find_update_fields if update else self._find_field
         for field in self._exclude_fields:
             for path, _ in find_fields(field, doc):
-                # If we found a matching field in the original document,
-                # delete it.
+                # Delete each matching field in the original document.
                 temp_doc = doc
                 for p in path[:-1]:
                     temp_doc = temp_doc[p]
@@ -493,8 +492,7 @@ class OplogThread(threading.Thread):
         find_fields = self._find_update_fields if update else self._find_field
         for field in self._fields:
             for path, value in find_fields(field, doc):
-                # If we found a matching field in the original document,
-                # copy it.
+                # Copy each matching field in the original document.
                 temp_doc = new_doc
                 for p in path[:-1]:
                     temp_doc = temp_doc.setdefault(p, {})
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -443,7 +443,7 @@ class OplogThread(threading.Thread):
         notation, eg "a.b.c". Returns a list of tuples (path, field_value) or
         the empty list if the field is not present.
         """
-        def find_all_fields():
+        def find_partial_matches():
             for key in doc:
                 if len(key) > len(field):
                     # Handle case where field is a prefix of key, eg field is
@@ -467,9 +467,12 @@ class OplogThread(threading.Thread):
                         # Stop searching, it's not possible for any other
                         # keys in the update doc to match this field.
                         return
-        if field in doc:
+
+        try:
             return [([field], doc[field])]
-        return list(find_all_fields())
+        except KeyError:
+            # Field does not exactly match any key in the update doc.
+            return list(find_partial_matches())
 
     def _pop_excluded_fields(self, doc, update=False):
         # Remove all the fields that were passed in exclude_fields.
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -419,59 +419,74 @@ class OplogThread(threading.Thread):
         self.running = False
         threading.Thread.join(self)
 
+    @classmethod
+    def find_field(cls, field, doc):
+        """Find the field in the document which matches the given field.
+
+        The field may be in dot notation, eg "a.b.c". Yields a tuple
+        (path, field_value).
+        """
+        path = field.split('.')
+        try:
+            for key in path:
+                doc = doc[key]
+        except (KeyError, TypeError):
+            return
+        yield path, doc
+
+    @classmethod
+    def find_update_fields(cls, field, doc):
+        """Find the fields in the update document which match the given field.
+
+        Both the field and the top level keys in the doc may be in dot
+        notation, eg "a.b.c". Yields tuples (path, field_value).
+        """
+        if field in doc:
+            yield [field], doc[field]
+            return
+        for key in doc:
+            if len(key) > len(field):
+                # Handle case where field is 'a' and key is 'a.b'
+                if key.startswith(field) and key[len(field)] == '.':
+                    yield [key], doc[key]
+                    # Continue searching, there may be multiple matches.
+                    # For example, 'a' will match 'a.b' and 'a.c'.
+            elif len(key) < len(field):
+                # Handle case where field is 'a.b' and key is 'a'
+                if field.startswith(key) and field[len(key)] == '.':
+                    remaining = field[len(key) + 1:]
+                    match = list(cls.find_field(remaining, doc[key]))
+                    if match:
+                        path, value = match[0]
+                        path.insert(0, key)
+                        yield path, value
+                        return
+
     def _pop_excluded_fields(self, doc, update=False):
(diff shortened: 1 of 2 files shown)


mongo_connector/oplog_manager.py
```diff
@@ -419,41 +419,74 @@ def join(self):
...
+                    # For example, 'a' will match 'a.b' and 'a.c'.
+            elif len(key) < len(field):
+                # Handle case where field is 'a.b' and key is 'a'
+                if field.startswith(key) and field[len(key)] == '.':
+                    remaining = field[len(key) + 1:]
+                    match = list(cls.find_field(remaining, doc[key]))
+                    if match:
+                        path, value = match[0]
+                        path.insert(0, key)
+                        yield path, value
```
llvtt: It seems that we end up calling `list` on the results of these two helper methods all over the place. Perhaps these should return lists instead of being generators?

mongo_connector/oplog_manager.py
```diff
@@ -419,41 +419,74 @@ def join(self):
...
+                        return
```
ShaneHarvey: That wouldn't work quite right. It's nessecary to continue searching in a sub-document. The case this code is handling can be shown with this example:
`find_update_fields('a.b.c.d', {'a.b': {'c': {'d': 1}}})` 

In this call we'll match the `'a.b'` to the beginning of `'a.b.c.d'` and search the sub-document for `'c.d'`, eg `cls.find_field('c.d', doc['a.b'])`

The insert at the beginning of the list only happens once so I don't see it as a problem and string manipulation is unavoidable here.

mongo_connector/oplog_manager.py
```diff
@@ -419,41 +419,74 @@ def join(self):
...
+                doc = doc[key]
+        except (KeyError, TypeError):
+            return
+        yield path, doc
+
+    @classmethod
+    def find_update_fields(cls, field, doc):
+        """Find the fields in the update document which match the given field.
+
+        Both the field and the top level keys in the doc may be in dot
```
ShaneHarvey: An update may or may not contain dot-notation field. For example, consider the case where `fields` is `['a.b.c']`. Updates that filter_oplog_entry should match could look like:
- `{'$set': {'a.b.c': 1}}`
- `{'$set': {'a.b': {'c':1}}}`
- `{'$set': {'a': {'b': {'c':1}}}}`

mongo_connector/oplog_manager.py
```diff
@@ -419,41 +419,74 @@ def join(self):
         self.running = False
         threading.Thread.join(self)
 
-    def _pop_excluded_fields(self, doc):
+    @classmethod
+    def find_field(cls, field, doc):
```
ShaneHarvey: Done.

mongo_connector/oplog_manager.py
ShaneHarvey: I agree it's pretty clunky. I'll change to return lists instead.

tests/test_filter_fields.py
```diff
@@ -653,19 +675,44 @@ def check_nested(document, exclude_fields, filtered_document):
...
+
+        update = {'$set': {'a.b.1': 1, 'a.b.2': 2, 'b': 3}}
+        exclude_fields = ['a.b']
+        filtered_update = {'$set': {'b': 3}}
+        check_nested(update, exclude_fields, filtered_update, op='u')
+
+        update = {'$set': {'a.b': {'c': 3, 'd': 1}, 'a.e': 1}}
+        exclude_fields = ['a.b', 'a.e']
+        filtered_update = None
+        check_nested(update, exclude_fields, filtered_update, op='u')
```
ShaneHarvey: I think that exact test case (ignoring field names) is handled above:
```python
update = {'$set': {'a.b.1': 1, 'a.b.2': 2, 'b': 3}}
exclude_fields = ['a.b']
filtered_update = {'$set': {'b': 3}}
...
```

mongo_connector/oplog_manager.py
llvtt: It looks like we stop searching if `match` is true, though, since we return right after the `yield`. Keep in mind that this is happening for every field for every document in the oplog, so even minutia like string manipulation/list insertion can have a significant impact on performance here.

tests/test_filter_fields.py
llvtt: That's not exactly the same. I'm asking about a test case where exclude_fields includes a top-level field, but the update spec is updating a field more than two levels deep.

mongo_connector/oplog_manager.py
ShaneHarvey: Okay, I see that the return can be unindented since an update cannot contain overlapping fields, eg `{'$set': {'a.b': {'c': {'d': 1}}, 'a': 2}}` is invalid. This will short circuit searches in some cases. 

I suppose the insert can be avoided too by passing an extra argument to `find_field`.

ShaneHarvey: It’s not clear to me how to make this matching more efficient without adding a lot more complexity. If you have a better algorithm in mind I’d like to hear it.

I kept the `insert(0, …)` because I don’t think there’s a difference if the insert happens in `_find_update_fields` or via passing an extra argument to `_find_field`. Either way, a list is created with `field.split('.')` and then the top level key is added.

tests/test_filter_fields.py
ShaneHarvey: Added that test to the tests.

mongo_connector/oplog_manager.py
llvtt: Sorry to be a pain about this. Perhaps I'm misunderstanding exactly what's going on here; can't the code be simplified to the following?

```
if field.startswith(key) and field[len(key)] == '.':
    matched = cls._find_field(field, doc)
    if matched:
        yield matched[0]
    return
```

ShaneHarvey: No, because `_find_field` assumes the document does not have dot-notation keys.

ShaneHarvey: The example I gave above shows why your suggestion does not work.

llvtt: Oh! I see now. Thanks for bearing with me. In that case, this is probably the best we can do.

ShaneHarvey: It's worth noting that we only use the slower `_find_update_fields` when we need to: when the user specifies --fields or --exclude-fields and the oplog entry is an update.

ShaneHarvey:
Merged https://github.com/mongodb-labs/mongo-connector/commit/293e3d45d4def8920e980847e4af8a57382b097a

robertaistleitner:
@ShaneHarvey thanks for your work on making this thing merge-ready! :+1: 
Unfortunately didn't have time to do it on my own.

PR #542, opened 2016-09-30
Log the current replication lag every 30 seconds

ShaneHarvey:
Logging replication lag is very useful for determining how mongo-connector is performing.
This change adds a thread per replica set that periodically logs the current replication lag (either in seconds or number of oplog entries).

If the current thread is up to date the log message will be:
`2016-09-30 15:26:46,616 [INFO] mongo_connector.oplog_manager:69 - OplogThread for replica set 'source' is up to date with the oplog.`

If the current thread is behind the log message will be:
`2016-09-30 15:18:45,114 [INFO] mongo_connector.oplog_manager:59 - OplogThread for replica set 'source' is 81 seconds behind the oplog.`

or:
`2016-09-30 15:18:15,014 [INFO] mongo_connector.oplog_manager:65 - OplogThread for replica set 'source' is 36 entries behind the oplog.`

Possible improvements:
- Make the logging interval a configuration option.
- Clear up the logging statements.
- Add more info: include the raw timestamps?

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -41,7 +41,7 @@ LOG = logging.getLogger(__name__)
 class ReplicationLagLogger(threading.Thread):
     """Thread that periodically logs the current replication lag.
     """
-    def __init__(self, opman, interval=30):
+    def __init__(self, opman, interval):
         super(ReplicationLagLogger, self).__init__()
         self.opman = opman
         self.interval = interval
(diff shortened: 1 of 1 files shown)


apalchys:
great! It will really help to understand what's going on and what's the status of mongo-connector.

mongo_connector/oplog_manager.py
```diff
@@ -38,6 +38,46 @@
...
 
 
+class ReplicationLagLogger(threading.Thread):
+    """Thread that periodically logs the current replication lag.
+    """
+    def __init__(self, opman, interval=30):
+        super(ReplicationLagLogger, self).__init__()
+        self.opman = opman
+        self.interval = interval
+        self.daemon = True
```
llvtt: You could pass this in to the super() constructor above.

ShaneHarvey: `daemon` was only [added as an argument to __init__ in Python 3.3](https://docs.python.org/3.5/library/threading.html#threading.Thread) so we should keep it as is for Python 2 compatibility.

Issue #514, opened 2016-08-25
incremental sync


kubage88:
when i restart mongo-connector process, it will not push data to elasticsearch any more, unless deleting oplog.timestamp, but if i do this , mongo-connector will sync data from beginning, i have a large amount  of data in mongodb,  how can i synchronize data incrementally when i restart mongo-connector process?

kubage88:
could you give me an answer? thanks a lot...

ShaneHarvey:
Sorry for the delay in getting back to you @kubage88. Mongo-connector is designed to handle a restart as long as the the `oplog.timestamp` oplog entry still exists in the oplog. If that happens mongo-connector logs an error message and shuts down. In the log do you see this error message `Last entry no longer in oplog cannot recover!`? If not this may be a bug. 

What version of mongo-connector and elastic-doc-manager are you using? I just tested your scenario with mongo-connector version 2.4.1 and elastic2-doc-manager version 0.2.0 and the connector was able to continue pushing updates to elastic after the restart.

kubage88:
you are so nice!
my mongo-connector's  version is  2.4.1 and elastic2-doc-manager's  version is 0.2.0。
my scenario is that when mongo-connnector  is pushing data to elasticsearch,  i press **ctrl+c** to stop the process manually,   then i run  **mongo-connector  -c  es.json**  in cmd again,   it will not push data  any more unless deleting oplog.timestamp.

the first zip file is the oplog when i pressed   **ctrl+c**.
the second zip file is the oplog when i ran    **mongo-connector  -c  es.json**     in cmd again for five minutes.

i hope it will be useful for you!

kubage88:
[log1.zip](https://github.com/mongodb-labs/mongo-connector/files/443961/log1.zip)
[log2.zip](https://github.com/mongodb-labs/mongo-connector/files/443962/log2.zip)
[es.json.zip](https://github.com/mongodb-labs/mongo-connector/files/443967/es.json.zip)

ShaneHarvey:
I was able to reproduce your issue. It seems that you are stopping the connector while it is performing the initial collection dump. The problem is that the oplog.progress file is already created but not all the documents have been synced to the target.

The giveaway was that your mongo-connector.log has many errors like:
```python
[...]

PR #543, opened 2016-09-30
Cancellable (not resumable) collection dump

ShaneHarvey:
Do not set the checkpoint when the initial collection dump is cancelled.
This means that on the next restart after an interrupted collection dump
mongo-connector will re-dump each collection.

Fixes the bug in #514.

Also, adds some extra logging in `dump_collection`.

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -659,7 +659,6 @@ class OplogThread(threading.Thread):
                     time.sleep(1)
 
         def upsert_each(dm):
-            num_inserted = 0
             num_failed = 0
             for namespace in dump_set:
                 from_coll = self.get_collection(namespace)
@@ -670,7 +669,6 @@ class OplogThread(threading.Thread):
                         mapped_ns = self.dest_mapping_stru.get(namespace,
                                                                namespace)
                         dm.upsert(doc, mapped_ns, long_ts)
-                        num_inserted += 1
                     except Exception:
                         if self.continue_on_error:
                             LOG.exception(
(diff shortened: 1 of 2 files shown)


mongo_connector/oplog_manager.py
```diff
@@ -668,16 +678,26 @@ def upsert_each(dm):
                             num_failed += 1
                         else:
                             raise
+                    if num % 10000 == 0:
+                        LOG.info("Upserted %d/%d docs from collection %s",
+                                 num + 1, total_docs, namespace)
```
llvtt: This isn't reliable, since documents may be inserted/removed while this is happening, changing the total. This could be confusing if the last line logged here is "Upserted 90/200 docs from collection %s," since it looks like mongo-connector may have missed a large portion of the collection. Clarifying that this is approximate could make this better.

mongo_connector/oplog_manager.py
```diff
@@ -668,16 +678,26 @@ def upsert_each(dm):
...
+                if num is not None:
+                    LOG.info("Upserted %d/%d docs from collection %s",
+                             num + 1, total_docs, namespace)
             LOG.debug("Upserted %d docs" % num_inserted)
```
llvtt: Seems like we're logging a lot of the same things here.

ShaneHarvey: Yeah. The total `num_inserted` can probably be removed now that we log per collection.

mongo_connector/oplog_manager.py
ShaneHarvey: Good point. Since there's nothing we can do to make it more accurate how's this:
```python
LOG.info("Upserted %d out of approximately %d docs from collection %s",
         num + 1, total_docs, namespace)
...
LOG.info("Bulk upserting approximately %d docs from collection %s",
         total_docs, namespace)
```

llvtt: Yeah, that looks good to me.

Issue #544, opened 2016-10-01
OplogThread should checkpoint skipped oplog entries


ShaneHarvey:
OplogThread currently does not update its checkpoint when skipping oplog entries. At best, this leads to re-processing these skipped oplog entries. At worst, the oplog may rollover with skipped entries causing the connector to fail even though no updates that matter happened.

Going even further, we may want to stop [filtering out no-ops from our oplog cursor](https://github.com/mongodb-labs/mongo-connector/blob/b3b4ae3199d2a341b15a56b6e86080ccc4021867/mongo_connector/oplog_manager.py#L547). It's possible that no-ops may cause the oplog checkpoint to rollover too. The majority of the changes would be updating the test suite to pass in the presence of no-ops.

PR #545, opened 2016-10-03
Update checkpoint even on skipped oplog entries

ShaneHarvey:
Resolves the first half of #544 but does not address filtering no-op oplog entries in `get_oplog_cursor`.

Also streamlines `update_checkpoint` so that:
```
self.checkpoint = timestamp
self.update_checkpoint()
```
becomes:
```
self.update_checkpoint(timestamp)
```

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -256,6 +256,70 @@ class OplogThread(threading.Thread):
             for ens in self.ex_namespace_set:
                 self._oplog_ex_ns_set.append(ens)
 
+    def _should_skip_entry(self, entry):
+        """Determine if this oplog entry should be skipped.
+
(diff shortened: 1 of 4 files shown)


mongo_connector/oplog_manager.py
```diff
@@ -380,16 +399,16 @@ def run(self):
...
+                            last_ts = None
 
                     # update timestamp after running through oplog
                     if last_ts is not None:
                         LOG.debug("OplogThread: updating checkpoint after"
                                   "processing new oplog entries")
-                        self.checkpoint = last_ts
-                        self.update_checkpoint()
+                        self.update_checkpoint(last_ts)
+                        last_ts = None
```
llvtt: I'm not sure this is necessary.

ShaneHarvey: Removed.

PR #570, opened 2016-10-27
Replace add_option with oplog_replay parameter

ShaneHarvey:
Small code clean up. No functional change and tested locally so I'm canceling the tests.

### DIFF
--- a/mongo_connector/oplog_manager.py
+++ b/mongo_connector/oplog_manager.py
@@ -608,9 +608,8 @@ class OplogThread(threading.Thread):
             query['ts'] = {'$gte': timestamp}
             cursor = self.oplog.find(
                 query,
-                cursor_type=CursorType.TAILABLE_AWAIT)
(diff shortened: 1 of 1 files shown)


mongo_connector/oplog_manager.py
```diff
@@ -608,9 +608,8 @@ def get_oplog_cursor(self, timestamp=None):
             query['ts'] = {'$gte': timestamp}
             cursor = self.oplog.find(
                 query,
-                cursor_type=CursorType.TAILABLE_AWAIT)
-            # Applying 8 as the mask to the cursor enables OplogReplay
-            cursor.add_option(8)
+                cursor_type=CursorType.TAILABLE_AWAIT,
+                oplog_replay=True)
```
llvtt: This option was added in PyMongo 3.0, but mongo-connector requires PyMongo >= 2.9, so it's possible this option won't be supported.

PR #575, opened 2016-10-28
Log startup info including versions of mongo-connector, Python, OS, and DocManagers

ShaneHarvey:
Add the idea of a `version` to a DocManager class. Downstream DocManager implementations should add their package version to the class similar to what is done by MongoDB DocManager in this pull request.
Change the default logging `verbosity` to 1 (warnings and above).
Add `LOG.always` which is always logged no matter the verbosity.

### DIFF
--- a/mongo_connector/__init__.py
+++ b/mongo_connector/__init__.py
@@ -11,20 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-import logging
-
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
-
-
-# Monkey patch logging to add Logger.always
-ALWAYS = logging.CRITICAL + 10
-
-logging.addLevelName(ALWAYS, 'ALWAYS')
-
-
-def always(self, message, *args, **kwargs):
-    self.log(ALWAYS, message, *args, **kwargs)
-
-logging.Logger.always = always
--- a/mongo_connector/__init__.py
+++ b/mongo_connector/__init__.py
@@ -17,17 +17,18 @@ import logging
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
+
 version_info = (2, 5, 0, 'dev0')
-version = '.'.join(str(v) for v in version_info)
-__version__ = version
+__version__ = '.'.join(str(v) for v in version_info)
 
 
 # Monkey patch logging to add Logger.always
-ALWAYS = 100
+ALWAYS = logging.CRITICAL + 10
 
 logging.addLevelName(ALWAYS, 'ALWAYS')
 
+
 def always(self, message, *args, **kwargs):
-    self._log(ALWAYS, message, args, **kwargs)
+    self.log(ALWAYS, message, *args, **kwargs)
 
 logging.Logger.always = always
--- a/mongo_connector/__init__.py
+++ b/mongo_connector/__init__.py
@@ -11,5 +11,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import logging
+
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
+
+version_info = (2, 5, 0, 'dev0')
+version = '.'.join(str(v) for v in version_info)
+__version__ = version
+
+
+# Monkey patch logging to add Logger.always
+ALWAYS = 100
+
+logging.addLevelName(ALWAYS, 'ALWAYS')
+
+def always(self, message, *args, **kwargs):
+    self._log(ALWAYS, message, args, **kwargs)
+
+logging.Logger.always = always
--- a/mongo_connector/__init__.py
+++ b/mongo_connector/__init__.py
@@ -18,10 +18,6 @@ from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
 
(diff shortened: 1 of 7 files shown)


mongo_connector/doc_managers/doc_manager_base.py
```diff
@@ -20,6 +20,10 @@
 class DocManagerBase(object):
     """Base class for all DocManager implementations."""
 
+    version_info = tuple()
+    version = 'unkown'
```
llvtt: This is somewhat counter to convention. Usually, the version is called `__version__` and is defined in a central module like `__init__.py`. Having it here in the class is somewhat strange. For example, the Elastic doc manager may have two classes, the DocManager and the BulkBuffer, but both of these classes conceptually have the same version.

Perhaps the convention could be to have a `__version__` defined at the module level. We can check for that when importing the doc manager in connector.py. Also, rather than having the string 'unknown' as a class member, we can just say 'unknown' if `not hasattr(module, '__version__')`.

mongo_connector/__init__.py
```diff
@@ -11,7 +11,23 @@
...
+
+import logging
+
 from pkgutil import extend_path
 __path__ = extend_path(__path__, __name__)
 
-__version__ = '2.5.0.dev0'
+version_info = (2, 5, 0, 'dev0')
+version = '.'.join(str(v) for v in version_info)
+__version__ = version
```
llvtt: Why both `version` and `__version__`? I think I've seen `__version__` more often in projects.

mongo_connector/connector.py
```diff
@@ -1103,7 +1104,7 @@ def setup_logging(conf):
             interval=conf['logging.rotationInterval'],
             backupCount=conf['logging.rotationBackups']
         )
-        print("Logging to %s." % conf['logging.filename'])
+        print("Logging to %s." % os.path.abspath(conf['logging.filename']))
```
llvtt: Is this consistently converted to an absolute path in other places? Perhaps this should be converted to an absolute path when parsing the config file/CLI options?

mongo_connector/doc_managers/doc_manager_base.py
ShaneHarvey: Changed to be defined at the DocManager implementation level. We can't use the `__init__.py` files as explained above. DocManager versions should probably be put on hold until we figure out what to do about the bigger problem.

mongo_connector/__init__.py
ShaneHarvey: So I just realized there's a larger problem here which is that `mongo-connector` and the docManagers all share the same `mongo-connector` module.  This means that if you do:
```
pip install mongo-connector
pip install elastic2-doc-manager
```
 the `__init__py` files of `elastic2-doc-manager` will clobber the existing ones (`mongo_connector/__init__.py` and `mongo_connector/doc_managers/__init__.py`). This means we can't rely on anything inside these files. In my opinion, this should be changed so that third party doc managers live in separate modules. Perhaps I should add versions in another file until that is settled.

llvtt: Can you not still import a `__version__` from the `mongo_connector` module? I just tried installing mongo-connector and the elastic-doc-manager side-by-side, and it looks like this still works. This could still be a problem if some doc manager tries to put its own version in `mongo_connector/__init__.py`, but doc manager authors shouldn't do that or need to.

I don't think we need to add another way to import doc managers. We already have two: use the `extend_path` trick or use the `docManagerClassPath` option in the config file (which circumvents this issue). If we stuck `__version__` in the doc manager module itself, instead of in `__init__.py`, we should be able to import the version from the doc manager without worrying about anything that might be in `mongo_connector/doc_managers/__init__.py`. If we can't import `__version__` reliably from mongo_connector, then maybe we need to put it back in constants.py.

ShaneHarvey: The reason it worked is because I moved `__version__` in #582. I'll rebase this PR and move the logging stuff out of `__init__.py` too. Yeah, we don't need to change the DocManager structure but need to be aware that they all share the same directory when installed so they can't have any files with the same name and `__init.py` should never be modified.

mongo_connector/__init__.py
```diff
@@ -17,17 +17,18 @@
...
+
 version_info = (2, 5, 0, 'dev0')
-version = '.'.join(str(v) for v in version_info)
-__version__ = version
+__version__ = '.'.join(str(v) for v in version_info)
 
 
 # Monkey patch logging to add Logger.always
-ALWAYS = 100
+ALWAYS = logging.CRITICAL + 10
```
llvtt: Nice.

mongo_connector/doc_managers/doc_manager_simulator.py
```diff
@@ -28,6 +28,18 @@
 from mongo_connector.doc_managers.doc_manager_base import DocManagerBase
 from mongo_connector.compat import u
 
+version_info = version_info
+__version__ = __version__
```
llvtt: Is this necessary?

mongo_connector/doc_managers/mongo_doc_manager.py
```diff
@@ -37,6 +37,18 @@
 
 LOG = logging.getLogger(__name__)
 
+version_info = version_info
+__version__ = __version__
```
llvtt: Same question

ShaneHarvey: Changed to be necessary.

mongo_connector/doc_managers/doc_manager_simulator.py
ShaneHarvey: Changed to be necessary.

ShaneHarvey:
Rebased and moved the logging.always setup out of `__init__.py`.

PR #582, opened 2016-11-01
Move __version__ out of __init__.py

ShaneHarvey:
`mongo-connector` and the docManagers all share the same `mongo-connector` module.  This means that if you do:
```
pip install mongo-connector
pip install elastic2-doc-manager
```
 the `__init__py` files of `elastic2-doc-manager` will clobber the existing ones (`mongo_connector/__init__.py` and `mongo_connector/doc_managers/__init__.py`). This means we can't rely on anything inside these files. In my opinion, this should be changed so that third party doc managers live in separate modules. Let's simply not add anything in these files for now.

### DIFF
--- a/mongo_connector/constants.py
+++ b/mongo_connector/constants.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+__version__ = '2.5.0.dev0'
+
 # Maximum # of documents to process before recording timestamp
 # default = -1 (no maximum)
 DEFAULT_BATCH_SIZE = -1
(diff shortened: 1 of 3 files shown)


mongo_connector/constants.py
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+__version__ = '2.5.0.dev0'
```
behackett: I don't think this is necessary. If a doc manager needed to compare versions it could use parse_version from pkg_resources.

behackett: Oops. I was commenting on the previous version.

Issue #461, opened 2016-05-26
Gracefully shutdown on SIGTERM and SIGINT


luisobo:
Hi there,

I'm not 100% sure but as far as I can tell, mongo-connector does not handle SIGSTOP. I'd expect it to listen to that signal and gracefully stop the threads, save the latest state to disk and exit. In SIGSTOP is not handled the only option is to use SIGKILL, which can potentially leave the system in an inconsistent state.

I'm using mongo-connector 2.3.0 with elastic2_doc_manager 0.1.0.

llvtt:
According to the man pages for BSD and Linux (http://www.unix.com/man-page/FreeBSD/3/signal/ and http://man7.org/linux/man-pages/man7/signal.7.html), SIGSTOP cannot be caught or ignored.

SIGKILL is the normal way to stop mongo-connector.

luisobo:
@llvtt you're 100% right. I meant SIGTERM, which is the signal that docker issues on docker stop. I edited the title, thanks. 

Regardless of the signal, the critical part is shutting down gracefully, which would include:
- closing connections
- killing threads
- writing progress file to disk
- exit with a status code of 0

Right now it takes a while for the system to shutdown. Seems that the system ignores SIGTERM (when running on docker) or SIGINT (when running in a console and doing Control+C) and only when it times out a SIGKILL is emitted. The process ends but with a status of 137 and I'm not sure if the progress file is properly persisted.

llvtt:
Oops, I got my signal wrong as well. I meant SIGINT is the normal way to stop the connector.

If the connector is not responding to a SIGINT within a minute or two, then something is wrong. Do you have the logs from mongo-connector available from a time when it wasn't responding?

PR #595, opened 2016-11-23
Gracefully exit on SIGTERM

ShaneHarvey:
Fixes https://github.com/mongodb-labs/mongo-connector/issues/461 and https://github.com/mongodb-labs/elastic2-doc-manager/issues/29.

Also, increases the time that `killproc` waits until sending `SIGKILL` from the default 5 seconds to 60 seconds.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Discovers the mongo cluster and starts the connector.
+"""Discovers the MongoDB cluster and starts the connector.
 """
 
 import json
@@ -22,6 +22,7 @@ import platform
 import pymongo
 import re
 import shutil
+import signal
 import ssl
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


sliwinski-milosz:
Hi,

I like this solution... but maybe it would be good to remove try/except and just handle in similar way SIGINT. Here is my proposition based on yours:
https://github.com/mongodb-labs/mongo-connector/compare/master...sliwinski-milosz:register_shutdown_handlers?expand=1

mongo_connector/connector.py
```diff
@@ -193,9 +197,9 @@ def join(self):
         """ Joins thread, stops it from running
         """
         self.can_run = False
+        threading.Thread.join(self)
```
llvtt: Can this use `super`?

mongo_connector/connector.py
```diff
@@ -1175,6 +1181,13 @@ def main():
     log_startup_info()
 
     connector = Connector.from_config(conf)
+
+    # register a SIGTERM handler to cleanup the connector gracefully
+    def sigterm_handler(signum, frame):
+        connector.signal = ('SIGTERM', signum)
```
llvtt: It doesn't look like we use `signum` anywhere. It would also be nice to handle SIGINT in the same way as SIGTERM, so we know we're handling cleanup/shutdown in the same way.

ShaneHarvey:
Thanks @sliwinski-milosz and @llvtt.

Issue #600, opened 2016-12-02
Namespace wildcard parsing is matching unwanted collections


letsgolesco:
### Background

My MongoDB has multiple databases with the same set of collections.

I'm using wildcards in the `namespaces.include` config option to transport the same collections from multiple databases, like so:
```json
  "namespaces": {
    "include": [
      "*.accounts"
    ]
  }
```

There is a collection called `accounts`, which I want to transport to Elasticsearch.

There is a collection called `deletedaccounts`, which I **do not** want to transport to Elasticsearch.

### Problem

Mongo Connector is transporting documents from the `deletedaccounts` MongoDB collection to Elasticsearch, creating a new Elasticsearch type called `delete.accounts`.

I found [this line in the dest_mapping.py file](https://github.com/mongodb-labs/mongo-connector/blob/293e3d45d4def8920e980847e4af8a57382b097a/mongo_connector/dest_mapping.py#L79) and it turns my `"*.accounts"` into the regex `/(.*).accounts/` which matches both `database.accounts` and `database.deletedaccounts`.

### Possible solution

A possible solution here is to escape regex characters in namespace strings (except for the `*` wildcard that we want to use in the regex). 

In my case, `/(.*).accounts/` would become `/(.*)\.accounts/`. 

I'm not sure if this solution would negatively effect other components.

ShaneHarvey:
Thanks for finding this bug! It should be fixed in https://github.com/mongodb-labs/mongo-connector/pull/601 by escaping special characters in the namespace, like "." and "$", before creating a regex like you suggested.

letsgolesco:
Thanks for getting on this quickly!

PR #601, opened 2016-12-02
Refactor DestMapping

ShaneHarvey:
DestMapping now includes command namespaces, eg including "db.col" should
also include "db.$cmd".
Improve regex matching by creating RegexObjects and caching lookups.
Update the config parsing with a new format for "namespaces.mapping". It's no longer required to repeat namespaces if you're using "namespaces.include" and "namespaces.mapping". Add a dictionary format which will be extended later to support additional fields per namespace (eg "includeFields"/"excludeFields"):

```json
{
 "namespaces": {
        "mapping": {
            "db.source1": "db.dest1",
            "db2.*": "db2.dest_*",
            "db.col_*": {
                "rename": "db.new_name_*"
            }
        }
}
```
Before this change such a config would look like:
```json
{
 "namespaces": {
        "include": ["db.source1", "db2.*", "db.col_*"],
        "mapping": {
            "db.source1": "db.dest1",
            "db2.*": "db2.dest_*",
            "db.col_*": "db.new_name_*"
        }
}
```

Fixes bugs with the original implementation:
- wildcards in database names no longer can match the '.' character
- escape metacharacters when converting to regex (fixes https://github.com/mongodb-labs/mongo-connector/issues/600).
- `map_db` and `unmap_namespace` now work with wildcards regardless if the
  original namespace was previously mapped already.
- OplogThread now calls `map_namespace` instead of `get`.
- Gridfs files should now be synced properly.
- In rollback logic, the namespace from the DocManager should not be re-mapped.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -123,8 +123,7 @@ class Connector(threading.Thread):
         # Replace the origin dest_mapping
         self.dest_mapping = DestMapping(
             kwargs.get('ns_set'), kwargs.get('ex_ns_set'),
-            kwargs.get('dest_mapping'), kwargs.get('fields'),
-            kwargs.get('exclude_fields'))
+            kwargs.get('dest_mapping'))
 
         # Initialize and set the command helper
         command_helper = CommandHelper(self.dest_mapping)
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -121,9 +121,10 @@ class Connector(threading.Thread):
         self.kwargs = kwargs
 
         # Replace the origin dest_mapping
-        self.dest_mapping = DestMapping(kwargs.get('ns_set', []),
-                                        kwargs.get('ex_ns_set', []),
-                                        kwargs.get('dest_mapping', {}))
+        self.dest_mapping = DestMapping(
+            kwargs.get('ns_set'), kwargs.get('ex_ns_set'),
+            kwargs.get('dest_mapping'), kwargs.get('fields'),
+            kwargs.get('exclude_fields'))
 
         # Initialize and set the command helper
         command_helper = CommandHelper(self.dest_mapping)
@@ -781,12 +782,20 @@ def get_config_options():
                 "Destination namespaces set should not"
                 " contain any duplicates.")
 
-        for key, value in dest_mapping.items():
-            if key.count("*") > 1 or value.count("*") > 1:
+        for source_name, value in dest_mapping.items():
+            # Mapping may be old style with only a target name, eg:
+            # "db.source": "db.dest",
+            # or new style which can support extra options per namespace:
+            # "db.source": {"rename": "db.dest", ...}
+            if isinstance(value, dict):
+                dest_name = value.get("rename", source_name)
+            else:
+                dest_name = value
+            if source_name.count("*") > 1 or dest_name.count("*") > 1:
                 raise errors.InvalidConfiguration(
                     "The namespace mapping source and destination "
                     "cannot contain more than one '*' character.")
-            if key.count("*") != value.count("*"):
+            if source_name.count("*") != dest_name.count("*"):
                 raise errors.InvalidConfiguration(
                     "The namespace mapping source and destination "
                     "must contain the same number of '*' characters.")
--- a/mongo_connector/dest_mapping.py
+++ b/mongo_connector/dest_mapping.py
@@ -124,7 +124,7 @@ class DestMapping(object):
         renames = {}
         for src_name, v in user_mapping.items():
             if isinstance(v, dict):
-                target_name = v.get('rename')
+                target_name = v.get('rename', src_name)
             else:
                 target_name = v
             renames[src_name] = target_name
@@ -238,8 +238,15 @@ class DestMapping(object):
     def map_db(self, plain_src_db):
         """Given a plain source database, return the list of target databases.
 
-        Individual collections in a database can be mapped to
-        different target databases, so map_db can return multiple databases.
+        Individual collections in a database can be mapped to different
+        target databases, so map_db can return multiple databases. This
+        function must return all target database names so we make the
+        following restrictions on wildcards:
+        1) A wildcard appearing in the source database name must also appear
+        in the target database name, eg "db*.col" => "new_db_*.new_col".
+        2) A wildcard appearing in the source collection name must also appear
+        in the target collection name, eg "db.col*" => "new_db.new_col*".
+
         This is used by the CommandHelper for the dropDatabase command.
         """
         if not self._regex_map and not self._plain:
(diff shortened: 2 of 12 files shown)


mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,223 @@
...
-            in the same time.
-            user_mapping should be non-empty together with namespace_set.
-            This assumption has been verified in the get_config_options()
-            in connector.py.
-        """
+
+class MappedNamespace(_MappedNamespace):
+    def __new__(cls, name=None, include_fields=None, exclude_fields=None):
+        return super(MappedNamespace, cls).__new__(
+            cls, name, include_fields, exclude_fields)
```
ShaneHarvey: include_fields/exclude_fields will be used in the next PR to add filtering fields per namespace.

mongo_connector/dest_mapping.py
```diff
@@ -162,8 +242,49 @@ def map_db(self, plain_src_db):
...
+    def fields(self, plain_src_ns):
+        """Get the fields to include and exclude for a given namespace."""
+        mapped = self.get(plain_src_ns)
+        if mapped:
+            return mapped.include_fields, mapped.exclude_fields
         else:
-            return [plain_src_db]
+            return None, None
+
+    def projection(self, plain_src_name, projection):
```
ShaneHarvey: Moving exclude_fields/include_fields/projection from OplogThread into DestMapping should probably be part of the next PR for filtering per namespace. The interaction of these fields here is kind of awkward -- I've already cleaned it up in the next change just forgot to remove it.

mongo_connector/oplog_manager.py
```diff
@@ -144,7 +134,7 @@ def __init__(self, primary_client, doc_managers,
 
         if not self.oplog.find_one():
             err_msg = 'OplogThread: No oplog for thread:'
-            LOG.warning('%s %s' % (err_msg, self.primary_connection))
+            LOG.warning('%s %s' % (err_msg, self.primary_client))
```
ShaneHarvey: This is a separate bug and could be moved out into it's own change.

mongo_connector/oplog_manager.py
```diff
@@ -229,33 +200,6 @@ def gridfs_files_set(self):
...
-
-    def update_oplog_ns_set(self):
-        self._oplog_ns_set = []
-        if self.namespace_set:
-            for ns in self.namespace_set:
-                self._oplog_ns_set.append(ns)
-                self._oplog_ns_set.append(ns.split('.', 1)[0] + '.$cmd')
-
-            self._oplog_ns_set.extend(self.gridfs_files_set)
-            self._oplog_ns_set.append("admin.$cmd")
```
ShaneHarvey: Why do we always add the "admin.$cmd" namespace?

mongo_connector/oplog_manager.py
```diff
@@ -291,28 +235,20 @@ def _should_skip_entry(self, entry):
...
-        if ns is None:
-            LOG.debug("OplogThread: Skipping oplog entry: "
-                      "'%s' is not in the namespace set." % (ns,))
-            return True, False
+        # Rename or filter out namespaces that are ignored keeping
+        # included gridfs namespaces.
+        new_ns = self.dest_mapping_stru.map_namespace(ns)
+        if new_ns is None:
+            if is_gridfs_file:
+                # Gridfs files should not be skipped
```
ShaneHarvey: I chose not to add the gridfs file set to DestMapping yet so it doesn't support wildcards. I don't have a good reason for why.

mongo_connector/dest_mapping.py
ShaneHarvey: Removed the unnecessary projection changes.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
-        self.ex_namespace_set = ex_namespace_set
+        self.namespace_set = set()
+        self.ex_namespace_set = RegexSet.from_namespaces(
+            ex_namespace_set or [])
 
         self.lock = threading.Lock()
 
+        user_mapping = user_mapping or {}
+        namespace_set = namespace_set or []
         # initialize
```
llvtt: Perhaps this comment could explain what the initialization does by default.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
 
-    def match_set(self, plain_src, dst_arr):
-        for x in dst_arr:
-            if plain_src == x:
-                return True
-            if "*" in x:
-                m = self.match(plain_src, x)
-                if m:
-                    return True
+    def _add_mapping(self, src_name, dest_name=None, include_fields=None,
```
llvtt: This might better be called `_add_collection` or `_add_collection_spec` since this is dealing with more than just the mapping now.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
+                "namespace mapping for: '%s'" % (src_name,))
+        if dest_name is None:
+            dest_name = src_name
+        self.set(MappedNamespace(dest_name=dest_name, source_name=src_name,
+                                 include_fields=include_fields,
+                                 exclude_fields=exclude_fields))
+        # Add the namespace for commands on this database
+        cmd_name = src_name.split('.', 1)[0] + '.$cmd'
+        dest_cmd_name = dest_name.split('.', 1)[0] + '.$cmd'
+        self.set(MappedNamespace(dest_name=dest_cmd_name, source_name=cmd_name))
```
llvtt: `set` is usually the name of a method that associates a value with a key; this looks like it's just adding something.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
 
-        return False
+    def set_plain(self, mapped_namespace):
+        """A utility function to set the corresponding plain variables"""
```
llvtt: I'm not sure what this docstring means.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
-                return None
+        # if plain_src_ns matches ex_namespace_set, ignore
+        if plain_src_ns in self.ex_namespace_set:
+            return None
+        if not self.regex_map and not self.plain:
+            # here we include all namespaces
+            return MappedNamespace(plain_src_ns)
+        # search in plain mappings first
+        try:
+            return self.plain[plain_src_ns]
```
llvtt: Does this need to be protected with the lock? It seems that we hold the lock elsewhere when we read/modify `self.plain`.

mongo_connector/dest_mapping.py
```diff
@@ -15,143 +15,230 @@
...
-            return None
-        # if no namespace_set, no renaming
-        if not self.namespace_set:
-            return self.get(plain_src_ns, defaultval=plain_src_ns)
+        mapped = self.get(plain_src_ns)
+        if mapped:
+            return mapped.dest_name
         else:
-            return self.get(plain_src_ns)
+            return None
```
llvtt: Looks like you don't need the `else` anymore.

mongo_connector/connector.py
```diff
@@ -781,12 +782,20 @@ def apply_namespaces(option, cli_values):
...
 
-        for key, value in dest_mapping.items():
-            if key.count("*") > 1 or value.count("*") > 1:
+        for source_name, value in dest_mapping.items():
+            # Mapping may be old style with only a target name, eg:
+            # "db.source": "db.dest",
+            # or new style which can support extra options per namespace:
+            # "db.source": {"rename": "db.dest", ...}
+            if isinstance(value, dict):
+                dest_name = value.get("rename", source_name)
```
llvtt: I like the second example. For simply including a collection, it may not matter what the value is (e.g. could be ``True``).

If we didn't mind changing the config file format a bit more, we could get rid of namespaces.(include|exclude) entirely like this:

```json
{
  "namespaces": {
    "included.collection": true,
    "excluded.collection": false,
    "renamed.collection": "something.else"
  }
}
```

mongo_connector/oplog_manager.py
llvtt: Huh, I don't know why I thought that...

PR #609, opened 2016-12-17
Add integration tests for renaming namespaces with the MongoDB DocManager

ShaneHarvey:
This change adds integration tests for MongoDB to MongoDB command replication on renamed namespaces and fixes the replication of such commands because of the typo in '$cmd' added here https://github.com/mongodb-labs/mongo-connector/pull/601/files#diff-9e5fc676a1f3b77e19871877266b4cf0R240.

Also, removes the `namespace_set` attribute from the MongoDB DocManager which is never initialized. `namespace_set` was added in version 1.2 https://github.com/mongodb-labs/mongo-connector/commit/bdfe9aad2db8b4adb6616b22fcc71a4e24feb26b but the initialization seems to have been unintentionally removed in version 2.0 https://github.com/mongodb-labs/mongo-connector/commit/6b8a6649e9008397ef2eb4fe1183541b90e2de5b#diff-ca51c2fabfa279cb1d2902ff6c2d8152L371

### DIFF
--- a/mongo_connector/doc_managers/mongo_doc_manager.py
+++ b/mongo_connector/doc_managers/mongo_doc_manager.py
@@ -110,17 +110,14 @@ class DocManager(DocManagerBase):
             return namespace
 
     @wrap_exceptions
-    def _namespaces(self):
-        """Provides the list of namespaces being replicated to MongoDB
+    def _meta_collections(self):
+        """Provides the meta collections currently being used
         """
         if self.use_single_meta_collection:
-            # This method will not return all the collections, only the ones
-            # with operations currently present in the capped meta collection.
-            # That is okay because _namespaces is only used to search for
-            # documents inside the meta collection.
-            return self.meta_database[self.meta_collection_name].distinct('ns')
+            yield self.meta_collection_name
         else:
-            return self.meta_database.collection_names()
+            for name in self.meta_database.collection_names():
+                yield name
 
     def stop(self):
         """Stops any running threads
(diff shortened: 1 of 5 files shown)


mongo_connector/doc_managers/mongo_doc_manager.py
```diff
@@ -114,9 +113,6 @@ def _get_meta_collection(self, namespace):
     def _namespaces(self):
         """Provides the list of namespaces being replicated to MongoDB
         """
-        if self.namespace_set:
-            return self.namespace_set
-
         user_namespaces = []
```
llvtt: It seems as though we use `_namespaces` only to find the collections on the "meta database" that we've created so far, for use in `get_last_doc` and `search`. This whole method could probably be simplified to return only the collection names on `self.meta_database`, instead of probing all databases & collections.

tests/test_mongo.py
```diff
@@ -103,18 +104,28 @@ def setUp(self):
...
-        self.conn.drop_database('test')
+        self.drop_all_databases()
 
         self.connector.start()
         assert_soon(lambda: len(self.connector.shard_set) > 0)
         assert_soon(lambda: sum(1 for _ in self._search()) == 0)
 
+    def drop_all_databases(self):
+        for name in self.mongo_conn.database_names():
+            if name != "local":
```
llvtt: Maybe add the 'admin' database here as well, since that would allow us to test with auth enabled. We haven't done that in awhile, but it should be possible.

mongo_connector/doc_managers/mongo_doc_manager.py
ShaneHarvey: Good catch! After digesting the code a bit more I think it makes more sense to return the meta collection names instead of the replicated collection names.

tests/test_mongo.py
ShaneHarvey: Done.

mongo_connector/doc_managers/mongo_doc_manager.py
```diff
@@ -111,24 +110,14 @@ def _get_meta_collection(self, namespace):
...
-                if coll.startswith("system"):
-                    continue
-                namespace = "%s.%s" % (database, coll)
-                user_namespaces.append(namespace)
-        return user_namespaces
+        if self.use_single_meta_collection:
+            yield self.meta_collection_name
+        else:
+            for name in self.meta_database.collection_names():
+                yield name
```
llvtt: Nice.

ShaneHarvey:
Merged.

PR #611, opened 2016-12-21
Fixed ssl_cert_reqs, especially when 'ignored'

makhdumi:
This is meant to fix the `sslCertificatePolicy` option, which is partly broken because:
- Setting `ssl.sslCertificatePolicy` in the config doesn't work. Only the CLI option is actually used
- Even if the CLI option is specified, it isn't passed to the `MongoClient` if it's set to `'ignored'`, because the value maps to `0` (`if ssl_cert_reqs:` doesn't happen)

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -112,13 +112,13 @@ class Connector(threading.Thread):
         ssl_keyfile = kwargs.pop('ssl_keyfile', None)
         ssl_cert_reqs = kwargs.pop('ssl_cert_reqs', None)
         self.ssl_kwargs = {}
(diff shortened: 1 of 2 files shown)


makhdumi:
I like the idea of new tests. I'll add them to this PR in the next day or so, thanks!

llvtt:
@makhdumi Any updates on this?

ShaneHarvey:
Merged in https://github.com/mongodb-labs/mongo-connector/commit/1d9d40cbb09b109a21b2a4c63466575d05e18c95, thanks @makhdumi!

Issue #621, opened 2017-01-02
Duplicated oplog entries with MongoDB 2.4 breaks mongo-connector


mikael-lindstrom:
I tried to use mongo-connector with MongoDB 2.4 and have found a problem with duplicated entries in the oplog. Basically $unset is written to the oplog even if the field being unset doesn't exist.

Example:
```
PRIMARY> db.version()
2.4.14
PRIMARY> use test;
switched to db test
PRIMARY> db.test.insert({"a":"1","b":"2"});
PRIMARY> db.test.find();
{ "_id" : ObjectId("586a4f39656f449ce6b62c25"), "a" : "1", "b" : "2" }
PRIMARY> db.test.update({"_id" : ObjectId("586a4f39656f449ce6b62c25")},{$unset:{"a":"1"}});
PRIMARY> db.test.find()
{ "_id" : ObjectId("586a4f39656f449ce6b62c25"), "b" : "2" }
PRIMARY> db.test.update({"_id" : ObjectId("586a4f39656f449ce6b62c25")},{$unset:{"a":"1"}});
PRIMARY> db.test.find()
{ "_id" : ObjectId("586a4f39656f449ce6b62c25"), "b" : "2" }
PRIMARY> use local
switched to db local
PRIMARY> db.oplog.rs.find();
{ "ts" : Timestamp(1483362083, 1), "h" : NumberLong(0), "v" : 2, "op" : "n", "ns" : "", "o" : { "msg" : "initiating set" } }
{ "ts" : Timestamp(1483362106, 1), "h" : NumberLong("-2373604974530658287"), "v" : 2, "op" : "i", "ns" : "test.test", "o" : { "_id" : ObjectId("586a4f39656f449ce6b62c25"), "a" : "1", "b" : "2" } }
{ "ts" : Timestamp(1483362127, 1), "h" : NumberLong("5823667433637975028"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : ObjectId("586a4f39656f449ce6b62c25") }, "o" : { "$unset" : { "a" : 1 } } }
{ "ts" : Timestamp(1483362141, 1), "h" : NumberLong("-1127311525069515859"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : ObjectId("586a4f39656f449ce6b62c25") }, "o" : { "$unset" : { "a" : 1 } } }
```

[...]

PR #631, opened 2017-01-21
Ignore KeyError for since the oplog can contain duplicates for Mongo

mikael-lindstrom:
MongoDB 2.4 can have duplicate `$unset` for a field in the oplog which causes mongo-connector to throw a `KeyError`. This fixes the problem by ignoring the KeyError for `$unset`. 

For more info, see #621


ShaneHarvey:
@mikael-lindstrom thanks! Merged https://github.com/mongodb-labs/mongo-connector/commit/f5f3cda5c27b2f74fe43d5697f5ff8825c3db5db.

PR #648, opened 2017-02-11
Fix MongoDB 2.4 $unset and $unset on array elements

ShaneHarvey:
This change builds upon #631 by adding test cases for the `DocManagerBase` class. I noticed that MongoDB 2.4 not only records $unsets on fields that don't exist but it doesn't do any kind of validation at all. Therefore, `apply_update` must ignore all errors (fields that don't exist, field access on an array, array out of bounds access, ...)  when attempting to $unset a field:

```javascript
> db.test.insert({_id:1, "a":[0,1,2,3,4,5]})
WriteResult({ "nInserted" : 1 })
> db.test.update({_id:1}, {$unset:{"not-present": true}})
WriteResult({ "nMatched" : 1, "nUpserted" : 0 })
> db.test.update({_id:1}, {$unset:{"a.1234.foo.bar": true}})
WriteResult({ "nMatched" : 1, "nUpserted" : 0 })
> db.test.update({_id:1}, {$unset:{"a.not-valid": true}})
WriteResult({ "nMatched" : 1, "nUpserted" : 0 })
> db.getSiblingDB('local').oplog.rs.find();
{ "ts" : Timestamp(1486772604, 1), "h" : NumberLong(0), "v" : 2, "op" : "n", "ns" : "", "o" : { "msg" : "initiating set" } }
{ "ts" : Timestamp(1486772652, 1), "h" : NumberLong("-2124585854495096815"), "v" : 2, "op" : "i", "ns" : "test.test", "o" : { "_id" : 1, "a" : [ 0, 1, 2, 3, 4, 5 ] } }
{ "ts" : Timestamp(1486772664, 1), "h" : NumberLong("7235944004452389876"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : 1 }, "o" : { "$unset" : { "not-present" : 1 } } }
{ "ts" : Timestamp(1486772685, 1), "h" : NumberLong("8315138122015249325"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : 1 }, "o" : { "$unset" : { "a.1234.foo.bar" : 1 } } }
{ "ts" : Timestamp(1486772693, 1), "h" : NumberLong("-4843035137332467192"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : 1 }, "o" : { "$unset" : { "a.not-valid" : 1 } } }
> db.test.find()
{ "_id" : 1, "a" : [ 0, 1, 2, 3, 4, 5 ] }
```

Also, fixes a separate bug. `apply_update` should not remove array elements that are $unset. The correct behavior is to set the element to `null`:

```javascript
> db.test.insert({_id:1, "a":[0,1,2,3,4,5]})
WriteResult({ "nInserted" : 1 })
> db.test.update({_id:1}, {$unset:{"a.3": true}})
WriteResult({ "nMatched" : 1, "nUpserted" : 0 })
> db.test.find()
{ "_id" : 1, "a" : [ 0, 1, 2, null, 4, 5 ] }
> db.getSiblingDB('local').oplog.rs.find();
{ "ts" : Timestamp(1486771255, 11), "h" : NumberLong(0), "v" : 2, "op" : "n", "ns" : "", "o" : { "msg" : "initiating set" } }
{ "ts" : Timestamp(1486772325, 1), "h" : NumberLong("5611170004231793160"), "v" : 2, "op" : "i", "ns" : "test.test", "o" : { "_id" : 1, "a" : [ 0, 1, 2, 3, 4, 5 ] } }
{ "ts" : Timestamp(1486772339, 1), "h" : NumberLong("5505252814820844457"), "v" : 2, "op" : "u", "ns" : "test.test", "o2" : { "_id" : 1 }, "o" : { "$unset" : { "a.3" : 1 } } }
```


mongo_connector/doc_managers/doc_manager_base.py
```diff
@@ -55,6 +55,37 @@ def _retrieve_path(container, path, create=False):
...
+                        where[index_or_key] = None
+                    else:
+                        # Unset field removes it entirely.
+                        del where[index_or_key]
+                else:
+                    del doc[to_unset]
+            except (KeyError, IndexError, ValueError):
+                # Ignore unset errors since MongoDB 2.4 records invalid
+                # $unsets in the oplog.
+                pass
```
llvtt: I'm a bit concerned that this could be masking legitimate problems. What if the destination has gone out of sync (e.g. rollback at destination) and we start applying updates with $unset to these documents? Before we would have logged an error, now we silently "apply" these changes, which could make recovery from this situation more difficult.

Perhaps we should consider if MongoDB 2.4 is worth supporting, since this behavior is incorrect, and this version has been end-of-lifed for awhile now. Alternatively, we could introduce yet another config option (*shudder*), or log this issue and continue (possibly unhelpful if we needed to recover from this situation).

ShaneHarvey: Perhaps we should drop 2.4 in the next feature release. For now, what if we only ignore these errors when the source is MongoDB 2.4?
```python
except (KeyError, IndexError, ValueError):
    # Ignore unset errors since MongoDB 2.4 records invalid
    # $unsets in the oplog.
    if source_mongodb_version >= 2.6:
        raise
```

mongo_connector/connector.py
```diff
@@ -61,6 +60,17 @@ def always(self, message, *args, **kwargs):
     'required': ssl.CERT_REQUIRED
 }
 
+_min_source_version = None
+"""The minimum MongoDB version in the source cluster."""
+
+
+def get_mininum_mongodb_version():
+    return _min_source_version
```
ShaneHarvey: Is there a better way to make the source MongoDB version visible in the DocManagers?

mongo_connector/connector.py
```diff
@@ -304,6 +314,8 @@ def run(self):
         """Discovers the mongo cluster and creates a thread for each primary.
         """
         self.main_conn = self.create_authed_client()
+        global _min_source_version
+        _min_source_version = Version.from_client(self.main_conn)
```
ShaneHarvey: I need to update this logic to find the minimum MongoDB version in the cluster since the DocManager is shared. Is there a nice way to run a command on all nodes in a cluster with PyMongo?

Also, should we periodically update the version throughout the lifetime of mongo-connector in case the source cluster if upgraded or downgraded?

mongo_connector/doc_managers/doc_manager_base.py
ShaneHarvey: It's worth noting that $unset "errors" are already ignored in the MongoDB DocManager and the Solr DocManager. In fact, the Solr DocManager already works around this problem https://github.com/mongodb-labs/solr-doc-manager/blob/cde7743/mongo_connector/doc_managers/solr_doc_manager.py#L217-L218.

mongo_connector/connector.py
ShaneHarvey: Updated to find the minimum replica set `mongod` version.

ShaneHarvey:
Done tweaking this. Any suggestions for improvements are welcome.

ShaneHarvey:
Rebased to fix merge conflicts. @llvtt @behackett thoughts? $unset errors are now only ignored when the source MongoDB version is < 2.6.

mongo_connector/doc_managers/doc_manager_base.py
```diff
@@ -55,6 +61,43 @@ def _retrieve_path(container, path, create=False):
...
+                        del where[index_or_key]
+                else:
+                    del doc[to_unset]
+            except (KeyError, IndexError, ValueError):
+                source_version = get_mininum_mongodb_version()
+                if source_version is None or source_version.at_least(2, 6):
+                    raise
+                # Ignore unset errors since MongoDB 2.4 records invalid
+                # $unsets in the oplog.
+                LOG.warning("Could not unset field %r from document %r."
```
llvtt: Space after `.`

behackett:
We're going to drop support for MongoDB 2.4 in the fall, likely coinciding with the release of MongoDB 3.6.

ShaneHarvey:
Merged https://github.com/mongodb-labs/mongo-connector/commit/53200e100f14a1820ec0c6cca6f2b72740979b44.

PR #655, opened 2017-02-15
Add tests for setting sslCertificatePolicy

ShaneHarvey:
Supersedes #611.

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -1075,10 +1075,20 @@ def get_config_options():
         if ssl_cert_reqs is None:
             ssl_cert_reqs = option.value.get('sslCertificatePolicy')
 
-        if ssl_cert_reqs is not None and ssl_cert_reqs not in _SSL_POLICY_MAP:
-            raise errors.InvalidConfiguration(
-                'sslCertificatePolicy (--ssl-certificate-policy) must be one '
-                'of %s, got "%s"' % (_SSL_POLICY_MAP.keys(), ssl_cert_reqs))
+        if ssl_cert_reqs is not None:
+            if ssl_cert_reqs not in _SSL_POLICY_MAP:
+                raise errors.InvalidConfiguration(
+                    'sslCertificatePolicy (--ssl-certificate-policy) must be '
+                    'one of %s, got "%s"' % (
+                        _SSL_POLICY_MAP.keys(), ssl_cert_reqs))
+            if pymongo.version_tuple < (3, 0) and ssl_cert_reqs != 'ignored' and not ssl_ca_certs:
+                raise errors.InvalidConfiguration(
+                    '--ssl-certificate-policy is not "ignored" and '
+                    '--ssl-ca-certs was not be provided. Either upgrade '
+                    'PyMongo to >= 3.0 to load system provided CA '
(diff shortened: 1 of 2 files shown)


mongo_connector/connector.py
```diff
@@ -1062,21 +1062,33 @@ def import_dm_by_path(path):
 
     def apply_ssl(option, cli_values):
         option.value = option.value or {}
-        ssl_certfile = cli_values.pop('ssl_certfile')
-        ssl_keyfile = cli_values.pop('ssl_keyfile')
-        ssl_cert_reqs = cli_values.pop('ssl_cert_reqs')
-        ssl_ca_certs = (
-            cli_values.pop('ssl_ca_certs') or option.value.get('sslCACerts'))
+        ssl_certfile = cli_values.get('ssl_certfile')
```
behackett: dict.get takes a "default" value, so you can do:

ssl_certfile = cli_values.get('ssl_certfile', cli_values.get('sslCertfile'))

tests/test_config.py
```diff
@@ -614,6 +630,33 @@ def test_client_options(self):
...
+            'oplogFile': from_here('lib', 'dummy.timestamp'),
+            'ssl': {
+                'sslCertfile': 'certfile.pem',
+                'sslKeyfile': 'certfile.key',
+                'sslCACerts': 'ca.pem'
+            },
+        }
+        for cert_policy, expected_ssl_cert_req in [
+                ('ignored',  ssl.CERT_NONE),
+                ('optional', ssl.CERT_OPTIONAL),
```
behackett: As a side note, CERT_OPTIONAL doesn't do anything for client connections. I've downplayed it in PyMongo's docs.

ShaneHarvey: The current docs say ssl_cert_reqs "must be one of the three values ssl.CERT_NONE (certificates ignored), ssl.CERT_REQUIRED (certificates required and validated), or ssl.CERT_OPTIONAL (the same as CERT_REQUIRED, unless the server was configured to use anonymous ciphers)."

That sounds like there's still a use-case for CERT_OPTIONAL.

mongo_connector/connector.py
ShaneHarvey: That does not work here because the cli_values dict always contains the parameter names.

tests/test_config.py
behackett: Though technically that is true, MongoDB doesn't support aNULL ciphers so it's a moot point.

In case you're wondering, CERT_OPTIONAL maps to SSL_VERIFY_PEER, which you can read about here:

https://www.openssl.org/docs/man1.0.1/ssl/SSL_CTX_set_verify.html

ShaneHarvey: Thanks, updated the description of sslCertificatePolicy "optional" to be more accurate.

ShaneHarvey:
Updated to be compatible with PyMongo 2.9. I think I've addressed all the comments.

mongo_connector/connector.py
behackett: Ah. Thanks for explaining.

behackett:
LGTM

ShaneHarvey:
Merged in https://github.com/mongodb-labs/mongo-connector/commit/c8851518cbec6753916d251dc84311ff2c2c01de.

PR #692, opened 2017-03-31
Add support of "bson.dbref.DBRef" data type for formatting value. The…

burnetjubian:
… "bson.dbref.DBRef" is used by Solr DBRef.


mongo_connector/doc_managers/formatters.py
```diff
@@ -96,6 +96,8 @@ def transform_value(self, value):
             return value
         elif isinstance(value, datetime.datetime):
             return value
+        elif isinstance(value, bson.dbref.DBRef):
+            return value.id
```
ShaneHarvey: What about the `DBRef.database` and `DBRef.collection`?

README.rst
```diff
@@ -1,3 +1,13 @@
+==============
+mongo-connector-solr
+==============
+This project is based on original mongo-connector::https://github.com/mongodb-labs/mongo-connector
+
+Change log:
+-----------
+Add support to DBRef whose datatype is "bson.dbref.DBRef".
+
+
```
ShaneHarvey: Remove this, the change-log is https://github.com/mongodb-labs/mongo-connector/blob/master/CHANGELOG.rst.

mongo_connector/doc_managers/formatters.py
burnetjubian: Yes, you are correct, this change does not cover **database** and **collection** value, but only the **id** value.
At least, it supports DBRef data from MongoDB, and it can be used by complex search like "join" in Solr.
In order to cover database and collection value, we need to modify function _transform_element_ and _format_document_ which I'm working on it.

README.rst
burnetjubian: I will remove this.

burnetjubian:
I have updated it to cover all database, collection and id data. I works perfect in my local environment. 
I am not good at Python, sorry if my code is repeating. I am very glad if someone can refine it.

burnetjubian:
BTW, have no idea why the CI is failed for MongoDB 2.4.14

KamalHr:
Does this work fine with elasticsearch?

Issue #728, opened 2017-08-15
Ability to control oplog progress file write


travisp:
Currently, the Connector writes to the oplog progress file at various points, but this is not controllable by the doc manager, and it's also not detectable by the doc manager. This falls down when using asynchronous sends or buffering, such as outputting to Kafka (unless we send and block after each message).

Instead, it would be helpful to be able to be able to receive a message in the doc manager when the progress file is about to be written to -- the doc manager could then block and ensure that all of its messages have been sent and acknowledged prior to saving oplog progress.

PR #729, opened 2017-08-16
Implement #728 Flush doc managers before oplog progress write

travisp:
This is my proposal for issue #728. Basically it gives us the ability to use an async or buffering doc manager (such as kafka) and ensure that writes are sent prior to writing oplog progress.

In our Kafka Doc Manager, the implementation essentially looks like:
```
def flush(self):
  self.producer.flush()
```

It is a different purpose than `commit()`, because with an async or buffering doc manager data can be lost without the flush()

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -241,6 +241,12 @@ class Connector(threading.Thread):
         if not items:
             return
 
+        # ensure DocManagers have completed operations before recording oplog
+        # progress
+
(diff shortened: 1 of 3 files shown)


makhdumi:
We also implemented something like this in our own branch. It's critical that the oplog progress is only updated after all docmanagers have committed. I'm not sure why it already isn't. It caused us significant data loss.

PR #737, opened 2017-08-30
Format UUID was set the same UUID into __id and UUID into document

akitoshka:
Format UUID was set the same for DBRef, UUID into__id and UUID into document.
Now there are two different approachs how the mongo-connector handles UUID.
The first approach looks like into elastic2_doc_manager, doc-manager sets UUID into Elasticsearch as is, using default str method (with '-' delimiters)
The second approach looks like as into formatter, mongo-connector get hex value (without delimiters).
So we have different behaviors if we work with __id or UUID into doc.
It is unexpected behavior so we have to fix this.

### DIFF
--- a/mongo_connector/doc_managers/formatters.py
+++ b/mongo_connector/doc_managers/formatters.py
@@ -87,7 +87,7 @@ class DefaultDocumentFormatter(DocumentFormatter):
             # Just include body of binary data without subtype
             return base64.b64encode(value).decode()
         elif isinstance(value, UUID):
-            return value.hex
+            return value
         elif isinstance(value, (int, long, float)):
             if isnan(value):
                 raise ValueError("nan")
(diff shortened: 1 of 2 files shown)


makhdumi:
This would destroy virtually everyone's existing setups. Someone please reject it.

You also modified it in the DefaultDocumentFormatter which is used by more than just the elastic2 docmanager.

For what it's worth, it isn't at all difficult to convert the hex string it's writing into whatever UUID format you need. IMO you should be doing that in your own code when reading from Elasticsearch, or maybe look into Elasticsearch computed fields.

The hex form of the UUID is its true, raw form, totally agnostic to language and what I personally expect and prefer. It is not unexpected behaviour.

ShaneHarvey:
@akitoshka I don't quite understand the problem you are trying to solve here. Can you provide an example?

PR #825, opened 2018-10-08
remove issues link from README

malekascha:
I noticed that there isn't an issues page anymore. If there's still a way to access it please let me know. When I click this link in the README it goes to the pull requests page, so I think it should be removed.


jaraco:
Not sure why the issues were disabled. I don't think it was intentional. Can you update the PR to re-point to the new /yougov home?

malekascha:
I updated all the links. Thanks for your help!

PR #830, opened 2018-10-13
Fade to black

jaraco:
Ran [black](/ambv/black) on the code base. The code is now correctly formatted.

(also fixes one SyntaxError on Python 3)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 88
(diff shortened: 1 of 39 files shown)


hugovk:
How about adding `black --check --diff` to the CI?

That would of course also require adding Python 3.6 or 3.7 to the CI, as Black needs 3.6+.

jaraco:
That's a good idea about adding the check, and also a good reminder about the required version. Due to the delays it will take to add that check (and its interactions with the rest of the suite), I'm going to defer that until later.

PR #835, opened 2018-10-15
Deprecate Python 2

jaraco:
This step is the first of two in dropping support for Python 2. Now mongo-connector should emit a warning when imported or started on Python 2.


behackett:
FYI - we've dropped Python 2.6 support from PyMongo in master. We don't have any plans to drop 2.7 support any time soon.

PR #902, opened 2020-08-03
Added support for "continueOnError" for GridFS dump

AndrewG10i:
Initial dump does not tolerate GridFS file insertion exception even when continueOnError set to true. This commit updates this behavior.


AndrewG10i:
Pull-request points to a wrong tag, will reissue now...

Issue #882, opened 2019-08-28
bson.errors.InvalidBSON still an issue even when pymongo has C extensions.


burnsjake:
In trying to run mongo-connector between two replicasets running mongo 3.6.14, it fails with the dreaded InvalidBSON error. There is nothing in the target replicaset. I have installed the mongo connector fresh and verified that pymongo has the C extensions.  I have also tried installing and pinning an old pymongo (2.9) and mongo-connector (2.3.0) to no avail.

This is a log from a run of mongo-connector (3.1.1) with pymongo (3.9.0):
2019-08-28 16:19:51,512 [ALWAYS] mongo_connector.connector:50 - Starting mongo-connector version: 3.1.1
2019-08-28 16:19:51,512 [ALWAYS] mongo_connector.connector:50 - Python version: 3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0]
2019-08-28 16:19:51,514 [ALWAYS] mongo_connector.connector:50 - Platform: Linux-5.0.0-25-lowlatency-x86_64-with-Ubuntu-19.04-disco
2019-08-28 16:19:51,514 [ALWAYS] mongo_connector.connector:50 - pymongo version: 3.9.0
2019-08-28 16:19:51,515 [WARNING] mongo_connector.connector:170 - MongoConnector: Can't find /home/ubuntu/venv/oplog.timestamp, attempting to create an empty progress log
2019-08-28 16:19:51,549 [ALWAYS] mongo_connector.connector:50 - Source MongoDB version: 3.6.14
2019-08-28 16:19:51,549 [ALWAYS] mongo_connector.connector:50 - Target DocManager: mongo_connector.doc_managers.mongo_doc_manager version: 3.1.1
2019-08-28 16:20:56,048 [CRITICAL] mongo_connector.oplog_manager:714 - Exception during collection dump
Traceback (most recent call last):
  File "/home/ubuntu/venv/mcon/lib/python3.7/site-packages/mongo_connector/oplog_manager.py", line 668, in do_dump
    upsert_all(dm)
  File "/home/ubuntu/venv/mcon/lib/python3.7/site-packages/mongo_connector/oplog_manager.py", line 651, in upsert_all
    dm.bulk_upsert(docs_to_dump(from_coll), mapped_ns, long_ts)
  File "/home/ubuntu/venv/mcon/lib/python3.7/site-packages/mongo_connector/util.py", line 33, in wrapped
    return f(*args, **kwargs)
[...]

PR #915, opened 2021-02-01
Added support for unicode_decode_error_handler BSON codec options

quirin-pauli:
We faced the following issue:
`bson.errors.InvalidBSON: 'utf-8' codec can't decode bytes in position 314-315: invalid continuation byte`

This pull request adds support to pymongo's BSON codec option to overwrite the default `unicode_decode_error_handler=strict'

### DIFF
--- a/mongo_connector/connector.py
+++ b/mongo_connector/connector.py
@@ -83,7 +83,7 @@ class Connector(threading.Thread):
     Creates, runs, and monitors an OplogThread for each replica set found.
     """
 
-    def __init__(self, mongo_address, doc_managers=None, **kwargs):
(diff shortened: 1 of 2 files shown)


howardkhl (approved):

howardkhl:
@quirin-pauli I made the changes to `connector.py`, set "unicodeDecode": "ignore" in `config.json`. However still encounters `bson.errors.InvalidBSON`

quirin-pauli:
@howardkhl this only ignores failures while decoding Unicode characters.

I just saw #882 where the `InvalidBSON` exception is raised but for another reason:
`bson.errors.InvalidBSON: year 46312 is out of range`

This pull request won't solve this issue and it looks like there's no easy solution to this problem.

PR #930, opened 2021-09-04
Support transaction oplog operations

AmrAdelKhalil:
**Summary**

As a common error inside the original repo  `mongo-connector` issue with sync transactions. This PR provides a change that could support this kind of transaction oplog.

**Key Changes**
* Change the way that the `run` function process the oplogs documents when it is a transaction oplog entry


AmrAdelKhalil:
See these: 
1- https://pypi.org/project/mongo-connector-transaction/
2- https://pypi.org/project/elastic2-doc-manager-transaction/

